PR #3, opened 2020-08-27
Added draft ADR for decision on whether to reuse Kubernetes components or implement orchestrator-agent from scratch

soenkeliebau:
Added draft ADR for decision on whether to reuse Kubernetes components or implement orchestrator-agent from scratch


ohessel (changes requested):
Good for me.
I only thought that reusing kubernetes operators would not necessarily mean also using kubernetes (but this might be even more complex)

adr/drafts/ADRx-defined_reuse_of_k8s.adoc
```diff
@@ -0,0 +1,118 @@
...
+** Oliver Hessel
+** Sönke Liebau
+* Date: 27.08.2020
+
+== Context and Problem Statement
+
+The Stackable project will need at least two components in order to orchestrate the software that gets deployed.
+These components are the orchestrator and the agents that manage bare-metal (or virtual) servers.
+
+During previous design discussions it became quite clear that a lot of the issues that we need to for these components are very similar to issues that Kubernetes faced an the past - and solved already.
```
lfrancke, line 20:
```suggestion
During previous design discussions it became quite clear that a lot of the issues that we need to for these components are very similar to issues that Kubernetes faced in the past - and solved already.
```

ohessel (approved):

PR #51, opened 2021-03-01
ADR009 Placement of services on nodes

lfrancke:


adr/ADR009-selector_support.adoc
```diff
@@ -0,0 +1,89 @@
+= ADR009: Placement of services on nodes
```
soenkeliebau, line 1: When I first saw this title I totally misunderstood it, as my first association was "where on the nodes should we put the services" - but that may just be me.

adr/ADR009-selector_support.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+* Status: accepted
+* Deciders:
+** Lars Francke
+* Date: 1.3.2021
+
+
+== Context and Problem Statement
+
+We need to decide on a syntax and feature set on how to place services on nodes.
```
soenkeliebau, line 14: With reference to my comment from above one or two additional sentences what exactly we are trying to solve here might make sense. Someone without the background knowledge of our discussions today may - at this point - struggle to understand this.

adr/ADR009-selector_support.adoc
```diff
@@ -0,0 +1,89 @@
+= ADR009: Placement of services on nodes
+Lars Francke <lars.francke@stackable.de>
+v0.1, 1.3.2021
+:status: draft
+
+* Status: accepted
```
soenkeliebau, line 6: I think, and I may be mistaken here, that I've always tried keeping the two status in sync..

adr/ADR009-selector_support.adoc
```diff
@@ -0,0 +1,89 @@
...
+* Deciders:
+** Lars Francke
```
soenkeliebau, line 8: Should we add Malte and myself here?

soenkeliebau (changes requested):

adr/ADR009-selector_support.adoc
```diff
@@ -0,0 +1,89 @@
...
+* Date: 1.3.2021
+
+
+== Context and Problem Statement
+
+We need to decide on a syntax and feature set on how to place services on nodes.
+
+In the Kubernetes world this is a bit easier as there are fewer shared resources, and the underlying node does not matter as much.
+The nodes are also often interchangable, and the identity of the nodes do not matter.
+For us it's often the case that users have dedicated hardware for this use-case and all machines should be used and at the same time there are often special nodes reserved for special tasks (e.g. more resilient machines for leader roles).
```
fwaibel, line 18: Following the discussion from yesterday: e.g. machines equipped with GPUs

adr/ADR009-selector_support.adoc
lfrancke: Do you have a suggestion for me? Happy to change this and also the title.

adr/ADR009-selector_support.adoc
lfrancke: I've now used "assign" 
From here: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/
Is that better?

adr/ADR009-selector_support.adoc
soenkeliebau: Maybe something along the lines of 

> Basically for every running process of a service the user needs to be able to somehow specify on which host(s) this should be run. This specification could be a regular expression over hostnames, exact matches on the hostname, some form of tagging, selecting via system properties, etc.

adr/ADR009-selector_support.adoc
```diff
@@ -68,11 +68,18 @@ We decided to go with the Kubernetes way of using label selectors.
...
 ----
 
-Here we see that for a node _type_ called `leader` we have an object with two top-level properties, one of them being `selectors`.
+* Here we see that for a node _type_ called `leader` we have an object with two top-level properties, one of them being `selectors`.
 `selectors` is again a map of names to objects.
-The names are chosen by the user and should describe the role group, often this will be just `default` or something similar but it allows to override configuration per role group while inheriting parent-level config.
+* The names are chosen by the user and should describe the role group, often this will be just `default` or something similar but it allows to override configuration per role group while inheriting parent-level config.
+
+In summary we decided that all our CRDs should follow a consistent pattern:
+* At the top level there is one property per role
```
soenkeliebau, line 76: This doesn't render correctly for me due to a missing blank line.

adr/ADR009-selector_support.adoc
```diff
@@ -68,11 +68,18 @@ We decided to go with the Kubernetes way of using label selectors.
...
+* Under each role there can be one or more role groups (in a `selectors` property)
+* Each role group has a single `selector` which is a Kubernetes `LabelSelector`
+* Each role group can have optional `instances` and `instancesPerNode` fields
```
soenkeliebau, line 79: Should we maybe explicitly mention that this is a operator specific example and might look different for every product?

soenkeliebau (changes requested):
One syntax issue I think, but content wise only a single comment/question

adr/ADR009-selector_support.adoc
lfrancke: Well, maybe I misunderstand you but my intention was to actually have this the same across all our operators?

soenkeliebau: In that case we actually had different understandings, yes. I thought we'd forego the roles for operators that don't need it, but I see the sense in having this consistent!

soenkeliebau (changes requested):
Changes look good to me, just one comment left open.

adr/ADR009-selector_support.adoc
lfrancke: I have chosen a different text but added something. Please take a look.

soenkeliebau (approved):

PR #52, opened 2021-03-01
ADR010 Command pattern

lfrancke:


soenkeliebau (changes requested):
You are going to hate me for this, but I really think that this would benefit from a little drawing to explain the chosen solution.

soenkeliebau (approved):

PR #53, opened 2021-03-12
Add ADR to define default paths used by OS packages and agent

soenkeliebau:
I've briefly documented what I believe should be the correct OS paths for some of the things we need to keep on disk.

I think this is correct according to https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html but have my doubts, whether it makes sense in all cases.

Specifically I wonder about 

1. /etc/opt/stackable - I have not often seen /etc/opt used for configuration.
2. Packages - do we want to keep these in /opt under our folder root as well?
3. Service configs, similar question, should those be somewhere else then /etc/opt/stackable/serviceconfig maybe a level up at /etc/opt/serviceconfig to clearly separate them from config for our components?

### DIFF
--- a/adr/ADR011-directory_structure.adoc
+++ b/adr/ADR011-directory_structure.adoc
@@ -6,41 +6,66 @@ v0.1, 12.03.2021
 * Status: accepted
 * Deciders:
 ** Sönke Liebau
-* Date:  12.03.2021
+** Lars Francke
+* Date:  17.03.2021
 
 == Context and Problem Statement
 
 To run the Stackable platforms some components will need to be installed on the managed servers.
-We will offer OS packages for some distributions, currently deb and RPM packages are available.
-To make the components behave the same we would like to agree on some default directories and what content we will put there.
+We will offer OS packages for a specified set of distributions, currently deb and RPM packages are available.
+To make the components behave the same regardless of the target platform, we would like to agree on some default directories and what content we will put there.
 
 This ADR is less to document alternative solutions, but rather to document the directory structure we agreed on, hence there are no alternative options discussed.
 Instead I have added excerpts from the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] to illustrate why I think the chosen paths are the correct ones.
 
+== Considered Options
+
+* Follow the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] and keep files under `/etc/opt` and `/var/opt`
+* Deviate from the Standard and remove the extra `opt` subdirectory
+
+== Decision Drivers
+* Predictability for users
+* Adherence to standards & best-practices
+
 == Decision
+We decided to deviate from what the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] specifies a little bit, in order to make the file location more familiar to our users.
+Please refer to the listing below for final locations.
+
+While the standard calls for variable data and config files to be located in `opt` subdirectories under `/var` and `/etc` we have never seen this actually being done in practice and would expect our users to be confused if they had to look for the files here.
+
+=== Benefits:
+
+* Config files should be easily located for the largest part of our users
+
+=== Drawbacks:
+
+* Strictly speaking we break the official requirements of where our config files should be - but we seem to be in good company, as no one really follows this recommendation anyway
 
 ----
 /
 ├── etc
-│   └── opt
-│       └── stackable
-│           ├── agent
-│           │   └── secure
-│           ├── serviceconfig
-│           └── zookeeper-operator
+│   └── stackable
+│       ├── agent
+│       │   └── secure
+│       ├── serviceconfig
+│       └── zookeeper-operator
 ├── opt
 │   └── stackable
 │       ├── agent
 │       ├── packages
 │       └── zookeeper-operator
 └── var
-    ├── log
+    ├── lib
     │   └── stackable
-    │       └── servicelogs
-    └── opt
+    │       └── agent
+    └── log
         └── stackable
-            └── agent
+            └── servicelogs
+                ├── namespace-product-name
+                └── namespace-product-name
 ----
+
+== Discussion of Options
 
 === Binaries for Stackable Components
 
@@ -51,7 +76,7 @@ In the example above there are subdirectories for the agent and the ZooKeeper op
 === Packages
 
 For worker servers which are under management by our agent, we will need to install packages that contain the upstream software that is being rolled out by the agent.
-As this is is also third party software but being used by Stackable we will install these under `/opt/stackable/' as well, but put them under a subdirectory `packages`
+As this is is also third party software but being used by Stackable we will install these under `/opt/stackable/' as well, but put them under a subdirectory `packages` to keep them separate from the Stackable software.
 
 This only refers to the default setting the agent is configured with though, this path can be freely chosen by the user.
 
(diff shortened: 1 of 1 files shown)


soenkeliebau:
I've updated this and it is now ready for review @lfrancke

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+* Date:  17.03.2021
+
+== Context and Problem Statement
+
+To run the Stackable platforms some components will need to be installed on the managed servers.
+We will offer OS packages for a specified set of distributions, currently deb and RPM packages are available.
+To make the components behave the same regardless of the target platform, we would like to agree on some default directories and what content we will put there.
+
+This ADR is less to document alternative solutions, but rather to document the directory structure we agreed on, hence there are no alternative options discussed.
+Instead I have added excerpts from the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] to illustrate why I think the chosen paths are the correct ones.
```
lfrancke, line 19:
```suggestion
Instead, I have added excerpts from the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] to illustrate why I think the chosen paths are the correct ones.
```

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+* Follow the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] and keep files under `/etc/opt` and `/var/opt`
+* Deviate from the Standard and remove the extra `opt` subdirectory
+
+== Decision Drivers
+* Predictability for users
+* Adherence to standards & best-practices
+
+== Decision
+We decided to deviate from what the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] specifies a little bit, in order to make the file location more familiar to our users.
```
lfrancke, line 31:
```suggestion
We decided to deviate from what the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] specifies, in order to make the file location more familiar to our users.
```

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+* Config files should be easily located for the largest part of our users
+
+=== Drawbacks:
+
+* Strictly speaking we break the official requirements of where our config files should be - but we seem to be in good company, as no one really follows this recommendation anyway
+
+----
+/
+├── etc
+│   └── stackable
```
lfrancke, line 47: There's NBSPs all over the place here. Intentionally?

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+=== Binaries for Stackable Components
+
+In accordance with https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#optAddonApplicationSoftwarePackages[3.13.1] we will install the binaries for our components under `/opt/stackable/<packagename>/binaryfile.
+
+In the example above there are subdirectories for the agent and the ZooKeeper operator which contain the executable files for these components.
+
+=== Packages
+
+For worker servers which are under management by our agent, we will need to install packages that contain the upstream software that is being rolled out by the agent.
+As this is is also third party software but being used by Stackable we will install these under `/opt/stackable/' as well, but put them under a subdirectory `packages` to keep them separate from the Stackable software.
```
lfrancke, line 79:
```suggestion
As this is also third party software but being used by Stackable we will install these under `/opt/stackable/' as well, but put them under a subdirectory `packages` to keep them separate from the Stackable software.
```

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+This only refers to the default setting the agent is configured with though, this path can be freely chosen by the user.
+
+=== Configuration Files for Stackable Components
+
+According to https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#etcoptConfigurationFilesForOpt[3.7.4], if our binaries are kept under `/opt` we should keep the associated config files in `/etc/opt` and replicate the same folder structure here that can be found in `/opt`.
+
+This default location seems like it is different from the industry standard, as we could not find any software that actually uses this path.
+Instead it is accepted best-practice to put configuration either directly under `/etc` or create a vendor subdirectory under `/etc` and keep configs there.
+
+Out of the current components, only the agent requires a config file that will be kept here.
```
lfrancke, line 90: I think this whole paragraph can go. It doesn't age well and it's not about "architecture" but a specific implementation.

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+The agent also requires TLS keys, these will by default be kept in a `/etc/opt/stackable/agent/secret` - but this setting can be overridden in the agent config.
+
+=== Service Configuration
+Following the logic from above, config files that are written for the actual services that are managed by Stackable (Apache Kafka, Apache Hadoop, ...) should also reside in `/etc/opt/stackable`.
```
lfrancke, line 94: "Following the logic from above" two times in a row but with two different conclusions. I know what you're getting at but it reads weird.

Maybe something like "As outlined above" oder "According to the standard"

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+Following the logic stated above, we will remove the `opt` from this path as well.
+
+It doesn't make sense to call the containing directory `packages`, like the parent directory for the binaries, so instead we have renamed it to `serviceconfig` to better express what it actually contains.
+
+This only refers to the default setting the agent is configured with though, this path can be freely chosen by the user.
```
lfrancke, line 100: Agent specifics again

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,118 @@
...
+
+=== Agent Working Directory
```
lfrancke, line 102: Again: I don't think we should/need to document specific things.

This could be about "working data" or just "data directories" or "variable data"

lfrancke (changes requested):

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,81 @@
...
+
+This ADR is less to document alternative solutions, but rather to document the directory structure we agreed on, hence there are no alternative options discussed.
+Instead I have added excerpts from the https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html[Filesystem Hierarchy Standard] to illustrate why I think the chosen paths are the correct ones.
+
+== Decision
+
+----
+/
+├── etc
+│   └── opt
```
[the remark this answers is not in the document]
soenkeliebau: That is on purpose to keep the directory tree looking right. I can also replace them if you think this is an issue..

adr/ADR011-directory_structure.adoc
```diff
@@ -0,0 +1,81 @@
...
+    │   └── stackable
+    │       └── servicelogs
+    └── opt
+        └── stackable
+            └── agent
+----
+
+=== Binaries for Stackable Components
+
+In accordance with https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#optAddonApplicationSoftwarePackages[3.13.1] we will install the binaries for our components under `/opt/stackable/<packagename>/binaryfile.
```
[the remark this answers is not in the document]
soenkeliebau: I've changed it to be the document title and the section number.

adr/ADR011-directory_structure.adoc
soenkeliebau: My intention for this ADR was actually to "tie down" as many specific paths as possible, so "specific implementation" was what I was going for :)

Sounds to me like we have a fundamentally different view in this case. I agree that the name "architectural" hints at only tackling overall architectural things in there, but I think there is value in having one central place where you can look up the paths used by our plattform.

soenkeliebau:
I have addressed a few comments, some are  still left open due to a fundamental question about what we want to use ADRs for .
I was planning to update this ADR as we go along and have it be the source of truth for all os paths that the Stackable Platform uses, hence also included specifics in it. 
@larsfrancke is of the opinion that we should keep this generic and leave implementation details out of it.

At the end of the day I am fine either way, but if we make this generic, then I think we should have a page detailing the paths somewhere else in the docs..

Anyone else have any thoughts on this?

adr/ADR011-directory_structure.adoc
lfrancke: I agree but that should - in my opinion - be the end-user documentation.

lfrancke (approved):

PR #58, opened 2021-04-29
Antora based documentation site

lfrancke:
See the result at https://docs.stackable.tech


soenkeliebau (approved):
LGTM (what I've looked at - which is nothing)

PR #62, opened 2021-04-30
Adds ZooKeeper to docs

lfrancke:


siegfriedweber (approved):
LGTM

A unified navigation instead of a navigation per component would be nice (see https://gitlab.com/antora/antora/-/issues/629).

lfrancke:
Thanks. Yes, the unified documentation is already on my radar.
Couchbase is using Antora: https://docs.couchbase.com/home/index.html
https://github.com/couchbase/docs-site/blob/master/antora-playbook.yml

PR #85, opened 2021-09-08
Fixed link to Quickstart documentation

Jimvin:
Replace the link: with xref: so that Antora builds the HTML version of the docs correctly.


soenkeliebau (approved):
LGTM
full disclosure: I've no idea about Antora, but it looks to be in line with what was done in other files

PR #88, opened 2021-09-30
Gitpod config

pipern:
This is the first `.gitpod.yml` file I've written, so very open to addressing concerns or recognizing mistakes.

The intention is that contributors can very easily preview the result of their changes, without needing any local tools or configuration at all. This will launch a remote IDE which includes automatically running the documentation build process and viewing the results.


local-antora-playbook.yml
```diff
@@ -0,0 +1,48 @@
...
+  url: https://docs.stackable.tech
+  start_page: home::index.adoc
+  robots: allow
+
+content:
+  sources: 
+  - url: https://github.com/stackabletech/agent.git
+    branches: main
+    start_path: docs
+  - url: /workspace/documentation
```
pipern, line 12: This is absolute path, rather than "./" due to https://gitlab.com/antora/antora/-/issues/329#note_356022771

.gitpod.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM gitpod/workspace-full:latest
+
+USER gitpod
+
+RUN npm i -g @antora/cli
+RUN npm i -g gulp
```
pipern, line 6: I'm not sure if/why this is needed; gulp seems to have to be installed globally AND locally?

https://stackoverflow.com/questions/22115400/why-do-we-need-to-install-gulp-globally-and-locally

soenkeliebau:
Hi @pipern - I will take a look at this, but it might end up being early next week. I have no experience with gitpod either, so will probably have to do some reading while looking at this.

lfrancke:
I'm with you @pipern but I believe this is better than nothing for now. I like the contribution.
Thank you both and sorry it took so long to review.

lfrancke (approved):

PR #89, opened 2021-10-01
Improve ubuntu installation path documentation

hikhvar:


modules/ROOT/pages/getting_started.adoc
```diff
@@ -175,7 +175,8 @@ You will need to configure the Stackable OS package repository on the worker nod
 ==== Debian and Ubuntu
     apt-get install gnupg openjdk-11-jdk curl
     apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 16dd12f5c7a6d76a
-    echo "deb https://repo.stackable.tech/repository/deb-dev buster main" > /etc/apt/sources.list.d/stackable.list
+    echo "deb [trusted=yes] https://repo.stackable.tech/repository/deb-dev buster main" > /etc/apt/sources.list.d/stackable.list
```
teozkr, line 178: `[trusted=yes]` shouldn't be required, since you install the GPG key right above.

teozkr (changes requested):
Hey, thanks for the PR! Ideally we shouldn't need the `[trusted=yes]` since that's a bit of a security regression (it disables the requirement for offline package signing). Apart from that, this LGTM.

modules/ROOT/pages/getting_started.adoc
hikhvar: ok, I had a problem on my hetzner VMs, since they complain about a missing InRelease file in the repository. Can't reproduce on my current test setup. So maybe it was a temporary problem?

hikhvar: Error message was:
```
E: The repository 'https://repo.stackable.tech/repository/deb-dev buster Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

teozkr: Yeah, sounds like that was a temporary issue then. I can't reproduce this on Debian 10 or Ubuntu 20.04.

hikhvar: removed this, PTAL.

teozkr (approved):

Issue #91, opened 2021-10-01
Onboarding papercuts


teozkr:
A (relative unorganized) bunch of issues I encountered when trying to set up a testing cluster by following https://docs.stackable.tech/home/getting_started.html:

- [x] Clarify supported/recommended distros (Debian 10, CentOS 7/8) (#96)
- [x] Curl is used before being installed (#96)
- [ ] Inconsistent CRD installation methods (copy/paste vs curl), consider adding an amalgamation YAML?
- [ ] Do we need to specify the repository with both the Repository k8s object and APT/YUM source lists?
- [ ] Do we need to install the JVM manually? (shouldn't the specific app packages depend on this?)
- [x] Deb packages and systemd commands are wrong (need to remove -server suffix) (#96)
- [ ] Deb convention is that packages automatically enable required systemd services (I believe this is different on RHEL/CentOS?)
- [x] Log notes should probably refer to journalctl rather than syslog (#96)
- [ ] Systemd services should implement sd-notify protocol so that systemd can notify the user immediately that startup failed
- [x] Kubeconfig also needs to be copied from /etc/rancher/k8s/k8s.yaml to /root/.kube/config on controller node (never documented) (#96)
- [x] Example ZookeeperCluster doesn't conform to CRD schema (seems to be an indentation error, works if I change it to https://gist.github.com/teozkr/4f373aed9e7d6c25a5d47fa290d8563b) (#96)
- [ ] Zookeeper operator keeps failing with `ERROR stackable_operator::controller: Reconciliation finished with an error, this should not happen, please file an issue err=QueueError { source: InitialListFailed { source: Api(ErrorResponse { status: "404 Not Found", message: "\"404 page not found\\n\"", reason: "Failed to parse error data", code: 404 }), backtrace: Backtrace(()) }, backtrace: Backtrace(()) }`
- [ ] Zookeeper CRD YAML doesn't include command CRDs
- [ ] Stackable Agent dies until manual restart if K8s API is unavailable (https://github.com/stackabletech/agent/issues/317)
- [x] Required ZooKeeper config fields aren't provided in example (`dataDir`, `initLimit`, `syncLImit`) (#96)
- [x] ZooKeeper fails to connect to peers, because they are configured to listen by hostname (which Debian binds to localhost by default) (#96)
- [ ] Operators don't mention which CRDs it's failing to find (https://github.com/kube-rs/kube-rs/issues/640)
[...]

PR #96, opened 2021-10-04
Start addressing papercuts

teozkr:
Addresses the docs-related parts of #91. Trying to stick to one commit per papercut, so that it's easy to back one of them out if it proves to be controversial.


Jimvin (approved):
LGTM

PR #105, opened 2021-10-28
Customer personas (first draft)

teozkr:
Based on what we came to in today's meeting (https://docs.google.com/document/d/1K6BIvARKWpum1XJ2rveYbJGK4s1cyOMxnuCj_QyjKyU/edit#) but cleaned up slightly.

Created a new internal section for this stuff. Probably want to move the ADRs into it as well at some point, but I'm not really sure about whether this is the correct way to manage it.


modules/internal/pages/personas.adoc
```diff
@@ -0,0 +1,158 @@
...
+Robot authentication::
+  Does the customer have an opinion on how servers authenticate each other? +
+  (For example: Kafka servers need to connect to ZooKeeper to maintain quorum, even if there are no Kafka clients at the time.)
+Network control::
+  DNS;;
+    Is the customer willing to delegate us a DNS zone, or are we "stuck" adapting to whatever they provision for us?
+  TLS;;
+    Does the customer want us to integrate into their existing TLS PKI? If so, is it automated? What are their issuing policies?
+  Kerberos;;
+    Does the customer want us to integrate into their existing Kerberos realm?
```
soenkeliebau, line 19: Maybe briefly mention trusts and if they are an option?

modules/internal/pages/personas.adoc
```diff
@@ -0,0 +1,158 @@
...
+  Overlay peering/BGP;;
+    Is the Kubernetes overlay network peered into their regular network environment?
+  NIC bonding & multihoming;;
+    Do they use bonded NICs that we need to take into account?
+  Network-level firewalls;;
+    Do they use network-level firewalls that our traffic needs to be allowed through?
+  Linux Security Modules?;;
+    Do they require all machines to use SELinux, AppArmor, or something similar?
+Kubernetess cluster::
```
soenkeliebau, line 28:
```suggestion
Kubernetes cluster::
```

modules/internal/pages/personas.adoc
```diff
@@ -0,0 +1,158 @@
...
+  Dedicated cluster;;
+    Can we expect to get a dedicated Stackable Kubernetes cluster, or will be be colocated with internal stuff?
+  Premises;;
+    Is it hosted in the cloud or on-premise?
+  Managed;;
+    Is it managed by an external service provider (IONOS, GKE, EKS, AKS, etc), by us, or is it managed by the customer themselves?
+  Node lifetime;;
+    How long can a single node identity be expected to last for? (Years, weeks, days, hours)
```
soenkeliebau, line 36: This mainly refers to the hostname of the node?

modules/internal/pages/personas.adoc
```diff
@@ -0,0 +1,158 @@
...
+  Kerberos usage;;
+    Issued manually by customer
+  Overlay peering/BGP;;
+    None
+  NIC bonding & multihoming;;
+    Yes
+  Network-level firewalls;;
+    Port-based, manually configured 
+  Linux Security Modules? (SELinux/AppArmor);;
+    ???
```
soenkeliebau, line 66: For all intents and purposes we can assume "no" here I believe. I've not yet seen anybody follow through on SeLinux..
@lfrancke  ?

soenkeliebau (changes requested):
Left some questions and one (nitpicking) correction. 

One general question, there is no distinction between inside/outside access to our services, we are assuming outside access is always necessary?

modules/internal/pages/personas.adoc
teozkr: Oops.

modules/internal/pages/personas.adoc
teozkr: Hostname, IP address, local disks, etc.

modules/internal/pages/personas.adoc
teozkr: I don't think I'm familiar enough to write a reasonable question here, but feel free to suggest something.

modules/internal/pages/personas.adoc
lfrancke: Agreed

teozkr: Ok, changed this to no then.

modules/internal/pages/personas.adoc
teozkr: Added a note about this to the doc

modules/internal/pages/personas.adoc
soenkeliebau, line 19:
```suggestion
    Does the customer want us to integrate into their existing Kerberos realm, if so is it an option to run our own kdc and integrate via trusts? 
```

soenkeliebau (approved):

soenkeliebau:
paging Mr. Antora (aka @lfrancke ) regarding the conflict in the antora part of the pr :)

lfrancke:
That's prerry straightforward. The antora.yml has changed. This just needs a manual resolve but that is easy.
We used to have multiple entries in there, now we only have one. Just add this new one at the end.

lfrancke (approved):

PR #107, opened 2021-11-04
added arch parameter

theredpenguin:
added arch parameter to repository otherwise a message is thrown that i386 packages are not found and the installation of stackable-packages is not working


siegfriedweber (approved):
Successfully tested.

Thank you for your contribution! I am happy to merge it as soon as you signed the CLA.

PR #117, opened 2021-11-17
ADR for authn token management

teozkr:
Rendered: https://github.com/stackabletech/documentation/blob/adr/authn-tokens/modules/adr/pages/drafts/ADRx-authn_token_management.adoc

This is a generalization of https://github.com/stackabletech/issues/issues/4, taking into account other secret types, as well as organizationally managed secrets.

Personally I'm leaning towards doing a MVP of the custom secret service. IMO, using Secrets directly won't work well for services that care about node identities (anything that's exposed outside of the cluster), and my feeling is that relying on separate operators for each kind of secret will end up with a very disjointed experience. Vault is way too ops-heavy to have a hard dependency on, and as far as I can tell their policy framework isn't powerful enough to express our needs either.

### DIFF
--- a/modules/adr/pages/ADR012-authn_token_management.adoc
+++ b/modules/adr/pages/ADR012-authn_token_management.adoc
@@ -1,17 +1,18 @@
-= Authentication token management
+= ADR012: Authentication token management
 Teo Klestrup Röijezon <teo.roijezon@stackable.de>
-v0.1, dd.mm.yyyy
-:status: draft
+v0.1, 2021-11-23
+:status: accepted
 
 * Status: {status}
-* Deciders: [list everyone involved in the decision] <!-- optional -->
-* Date: [YYYY-MM-DD when the decision was last updated] <!-- optional -->
+* Deciders:
+** Teo Klestrup Röijezon
+** Lars Francke
+** Sönke Liebau
+* Date: 2021-11-23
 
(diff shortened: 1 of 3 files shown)


lfrancke:
Thank you for preparing this. It is a great summary. We'll talk about it in our weekly architecture meeting.

modules/adr/pages/ADR012-authn_token_management.adoc
```diff
@@ -0,0 +1,78 @@
+= ADR012: Authentication token management
+Teo Klestrup Röijezon <teo.roijezon@stackable.de>
+v0.1, 2021-11-23
+:status: accepted
+
+* Status: {status}
+* Deciders:
```
teozkr, line 7: Should this include everyone who was present in the meeting, or only the people who were actively involved in the decision?

siegfriedweber: In my opinion, the actual decision makers are sufficient.

siegfriedweber (approved):
LGTM

modules/adr/pages/ADR012-authn_token_management.adoc
lfrancke: We copied the template from somewhere...I'm not super vested in any of the metadata or form :)

PR #134, opened 2022-02-04
ADR012 Supported Kubernetes Versions

fhennig:
https://github.com/stackabletech/issues/issues/113

I incoporated Sigis research into the ADR, added a bit of mine own too.

Maybe the reasoning behind the decision could be fleshed out more, I added some pros and cons how I thought about the decision.


fhennig:
I want to add that according to my understanding, it would be best that we from now on use the 3rd latest kube API to compile against, so at the moment that would be 1.20. Since all APIs in kube are always supported for at least 3 versions, that would - to my understanding - mean that our binaries are also compatible with the latest three versions. I think it might be as simple as changing the `v1_22` in [this line](https://github.com/stackabletech/operator-rs/blob/main/Cargo.toml#L18) to a `v1_20`

modules/adr/pages/ADR012-supported-kubernetes-versions.adoc
```diff
@@ -0,0 +1,108 @@
+= ADR012: Supported Kubernetes versions
+The Stackable Engineers
+v0.1, 04.02.2021
+:status: draft
+
+* Status: {status}
+* Deciders: Lars Franke
```
lfrancke, line 7:
```suggestion
* Deciders: Lars Francke
```

:)

modules/adr/pages/ADR012-supported-kubernetes-versions.adoc
```diff
@@ -0,0 +1,108 @@
...
+* Date: 2022-02-04
+
+Technical Story: [Document supported Kubernetes versions| https://github.com/stackabletech/documentation/issues/83]
```
lfrancke, line 10: Wrong link syntax

modules/adr/pages/ADR012-supported-kubernetes-versions.adoc
```diff
@@ -0,0 +1,108 @@
...
+Related Documentation: https://docs.stackable.tech/agent/limitations.html
```
lfrancke, line 11: This seems...irrelevant

modules/adr/pages/ADR012-supported-kubernetes-versions.adoc
```diff
@@ -0,0 +1,108 @@
...
+* "Rule #7: Deprecated behaviors must function for no less than 1 year after their announced deprecation."
+
+This means that a codebase relying on alpha features can only reliably be developed against a single kubernetes version. For beta features, 3 kubernetes versions can be covered with the same codebase. The linked document also shows the overlapping of versions. When a feature matures from beta into stable, the beta version can still be used for 3 releases. After that, the stable version has to be used.
+
+For actually deprecated features (not just moved from beta to stable), the grace period is also more than three releases; as specified a deprecated feature has to be supported for at least a year.
+
+=== Stackables behavior
+
+The Stackable services (agent and operators) support only a small subset of Kubernetes versions. The reason is that, as pointed out already, the kubernetes API and tools restrict support of backward compatibility. Our software is compiled against a specific Kubernetes version but run on all other versions if no breaking change is actually used.
```
lfrancke, line 35: Could you remove all references to the agent?

modules/adr/pages/ADR012-supported-kubernetes-versions.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+== What others support (September 2021)
+
+=== Kubernetes
+
+> The Kubernetes project maintains release branches for the most recent three minor releases (1.22, 1.21, 1.20). Kubernetes 1.19 and newer receive approximately 1 year of patch support. Kubernetes 1.18 and older received approximately 9 months of patch support.
+
+see https://kubernetes.io/releases/ and https://kubernetes.io/releases/version-skew-policy/
+
+=== Amazon EKS
```
lfrancke, line 49: Could you also check what IONOS, Azure and Google support?

I also don't think we need the "current" list but a link is definitely useful.

lfrancke (changes requested):

lfrancke (approved):

PR #136, opened 2022-02-08
Include all products in the supported versions list

### LINKED ISSUE #121: Document supported product versions
I'd like a page that summarizes for each operator the supported product versions.

Please see the epic at https://github.com/stackabletech/issues/issues/139 for more details.
Ideally the information can be aggregated from the operators themselves so the content is only written once.

teozkr:
Fixes #121

### DIFF
--- a/modules/ROOT/pages/supported-versions.adoc
+++ b/modules/ROOT/pages/supported-versions.adoc
@@ -2,10 +2,46 @@
 
 The latest versions of each of the Stackable Operators support the following product versions:
 
-== Apache ZooKeeper
+== Apache Airflow
 
-include::zookeeper::partial$supported-versions.adoc[]
+include::airflow::partial$supported-versions.adoc[]
+
+== Apache Druid
+
+include::druid::partial$supported-versions.adoc[]
+
+== Apache HBase
+
+include::hbase::partial$supported-versions.adoc[]
+
+== Apache HDFS
+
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/supported-versions.adoc
```diff
@@ -2,10 +2,46 @@
...
-include::zookeeper::partial$supported-versions.adoc[]
+include::airflow::partial$supported-versions.adoc[]
+
+== Apache Druid
+
+include::druid::partial$supported-versions.adoc[]
+
+== Apache HBase
+
+include::hbase::partial$supported-versions.adoc[]
```
teozkr, line 15: This depends on https://github.com/stackabletech/hbase-operator/commit/974c6e99aec068e47293ab83ae074674e8769f41 (from the rework branch), and will cause an ugly warning on the docs page until then.

That's not great, but it still seems like a better outcome long-term then forgetting to include HBase once the rework is merged.

maltesander (approved):
LGTM. Do we still want to wait for HBase to avoid that warning so far?

teozkr:
@lfrancke @siegfriedweber What do you think?

modules/ROOT/pages/supported-versions.adoc
lfrancke: Agreed!

lfrancke (approved):

PR #143, opened 2022-02-17
ADR draft: User authentication

teozkr:
Part of https://github.com/stackabletech/issues/issues/170. `AuthenticationClass` is a slightly minimized variant of the design that @Jimvin, @soenkeliebau, and I came up with, with the other two being intermediary steps that were also discussed at the time.


modules/contributor/pages/adr/drafts/ADRx-user_authentication.adoc
```diff
@@ -0,0 +1,147 @@
...
+[source,yaml]
+----
+apiVersion: nifi.stackable.tech/v1alpha1
+kind: NifiCluster
+metadata:
+  name: nifi
+spec:
+  authenticationConfig:
+    methods:
+    - authenticationClass: nifi-ldap-authn
```
fhennig, line 117: Is that `nifi-` prefix a typo or is it referencing an object that is not shown?

fhennig:
Hey! I like the ADR, I like the pro/con list for all the options, it makes sense to me. Just for the last option, I have difficulty understanding it. I thought about what to ask, but I can't pinpoint anything specific. Instead I want to ask: From inside the nifi operator, how would I get all the LDAP info I need to connect to LDAP? (And how does it work for - lets say - druid, is it the same?) This sentence confused me: " ... which allows secret-operator to bind in separate credentials for each cluster ...".

Some guesses I have: namespaces are probably important here (similarly to the discovery mechanism) but if they are, they are not mentioned. Multiple objects with the same name exist in different namespaces?

I hope it's clear what I'm asking :sweat_smile:

modules/contributor/pages/adr/drafts/ADRx-user_authentication.adoc
teozkr: That is a typo, and should just have been `ldap-authn`.

teozkr:
> From inside the nifi operator, how would I get all the LDAP info I need to connect to LDAP?

For public information: You fetch the `AuthenticationClass` object, and process that into an acceptable product-config setup for that product.

For secrets: you bind it as a regular secret-operator volume, and the NiFi operator never sees the actual values (it only knows "this password will be stored in the file /secrets/ldap-bind/password" (or whatever path we end up with)).

> Some guesses I have: namespaces are probably important here (similarly to the discovery mechanism) but if they are, they are not mentioned. Multiple objects with the same name exist in different namespaces?

Secret-Op already supports a concept named "scopes", which allows us to resolve these credentials on a number of different criteria. For example: per pod identity, per node, per namespace, and so on. https://github.com/stackabletech/secret-operator#usage tries to explain this a bit, but that's definitely an area that needs better docs.

fhennig:
Thank you for explaining, I think I understand it now. Just looking at it naively I am a bit unsure if the use of the secret-operator is necessary in this case. For a TLS certificate that needs to be issued a file seems sensible, and an operator that controls the CA as well. To me it looks like a basic secret would be enough in this case, but I could be wrong.

teozkr:
The benefits of routing it via Secret-Op are, to me:

1. It lets the cluster operator define the scope of the credentials (per node, per project, per cluster)
2. It's the ~only way to mount secrets across namespaces, for customers who want to use the same LDAP credentials for everything
3. It opens up the door for dynamic credentials (mTLS/Kerberos?) down the line

modules/contributor/pages/adr/drafts/ADRx-user_authentication.adoc
soenkeliebau: Ah, that clears up my main confusion as well I think!

sbernauer:
Closing in favor of https://github.com/stackabletech/documentation/pull/161

PR #146, opened 2022-02-23
ADR draft: product image versioning

teozkr:
Based on the discussion from today's architecture review meeting.


modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+=== Resolve SemVer ranges in operator
+
+The operator resolves a hard-coded SemVer range into a Docker tag and concrete image hash during reconciliation.
+
+* Good, because we preserve SemVer for image versions
+* Good, because we prevent version drift between cluster replicas
+* Good, because existing operators will upgrade as far as they are compatible
+* Good, because we can fail with a reasonable error message if no compatible image is available
+* Bad, because it increases the complexity of the operators even further
```
soenkeliebau, line 104: I have quite frankly stopped caring about this, if we can solve it properly once and manage to hide the complexity of resolving this in the framework ... so be it.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Bad, because it slows down reconciliation
```
soenkeliebau, line 105: Due to the external call to the registry?

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Bad, because existing operators will switch which (minor-level) image version they deploy without user action
```
soenkeliebau, line 106: Could we make this configurable by keeping the currently deployed image hash in the cluster status and allowing the user to configure whether the image version may be automatically updated?
Going a bit further, we could also track when we last checked for updates and allow defining a delay before the next check, that might help with delaying the reconciliation as well.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Bad, because we don't have a good trigger for when new image versions are released
```
soenkeliebau, line 107: Basically we'd just need to check (see above for ideas on how to mitigate the overhead a bit) on every reconciliation and hope that one is triggered every so often?

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Bad, because on-prem registry mirrors may not have whatever API or metadata we need to perform this resolution
```
soenkeliebau, line 108: And here I was hoping there'd be a common api for this ..

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: Yep

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: Falling back is an alternative, but there's a tricky balancing act here between stability, predictability and ensuring that we're running in a supported configuration.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: Either that, start doing interval requeues, or have an API somewhere for querying whether there's a new release available.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
lfrancke: You're clearly new to IT.
I have some more surprises for you then!

teozkr: Looks like [there is a common API](https://docs.docker.com/registry/spec/api/#listing-image-tags). I don't think it's strictly required for `docker pull` so I suspect not all registries support it... That said, [NXRM seems to](https://docker.stackable.tech/v2/stackable/kafka/tags/list), which is encouraging.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
soenkeliebau: Not sure I understand what do you mean by "falling back" in your comment?

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
soenkeliebau: yeah..

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
soenkeliebau: We could externalize this into a stackable operator that works across all services deployed by us, which performs a regular check and triggers a requeue for anything that has a new image available..

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
soenkeliebau: Would have been awesome if that api allowed restricting which tags to get, instead of just supporting "give me everything"..

Kinda makes one think of caching that response in a component we control somewhere..

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: Using a cached "current version" either rather than querying the registry, or because the registry is unavailable.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: I'm not sure how much we'd gain from that. I'd rather not teach one operator about all of our product CRD types (we've been through that dependency hell once already :P), and [Docker's Registry API](https://docs.docker.com/registry/spec/api/#detail) doesn't have an endpoint for "give me all tags for everything you know about", so we wouldn't save much network traffic by centralizing this into one component.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+=== Negative Consequences <!-- optional -->
+
+* [e.g., compromising quality attribute, follow-up decisions required, …]
+* …
+
+== Pros and Cons of the Options
+
+=== Synchronize image and operator versions
+
+Each operator version hard-codes exactly one image version that it supports, and we release new product images for each operator images.
```
siegfriedweber, line 61: Nightly builds of operators are not considered here. In #147, I proposed to use the image pull policy "Always" for nightly builds. This has the advantage that for the development the same mechanism is used (the operator 1.0.0-nightly pulls the product image tagged with 1.0.0-nightly) and that the development can progress because always the latest version of the product image is used. The disadvantages of the image pull policy only apply to the nightly versions but not to the releases. The only downside is that the nightly versions behave differently to the releases.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+This would likely mean moving the product Dockerfiles into their corresponding operators' repositories.
+
+* Good, because it centralizes the information about which versions are supported into each operator's repository
+* Good, because upgrades are predictable for the user, "upgrading the operator upgrades the cluster" is easy to explain and teach
+* Good, because image tags are stable and immutable once released
+* Bad, because we end up storing a lot of duplicate Docker images
+  * We could share the Docker layers to lessen this impact dramatically, but that would require rearchitecting our CI
+* Bad, because it increases the overhead of doing operator or image releases
```
siegfriedweber, line 70: Is this equivalent to "Bad because the operator must be released even if just the product image must be changed, e.g. due to a security update of the base image." in #147?

siegfriedweber (commented):
All differences between #147 and this pull request outlined.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+`docker.stackable.tech/stackable/$PRODUCT:$PRODUCT_VERSION-stackable$IMAGE_SEMVER_MAJOR`.
+
+In theory, a given Operator version depends on some SemVer range of `image_version` (for example `^1.3.0`, which
+expands to `>=1.3.0 <2.0.0`).
+
+In practice however, Docker tags are opaque, which prevents us from actually specifying this anywhere. Instead, we can only
+express the range `^N.0.0` (`>=N.0.0 <N+1.0.0`), depending on the `-stackable$IMAGE_SEMVER_MAJOR` tag. This works
+OK for spinning up new clusters, but means that existing clusters may keep running old image versions even after upgrading to new
+(incompatible) Operator versions. The cluster may also end up in a mixed environment if new nodes are added after a new image
+version is released.
```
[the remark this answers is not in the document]
teozkr: True.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+Each operator version hard-codes exactly one image version that it supports, and we release new product images for each operator images.
+
+This would likely mean moving the product Dockerfiles into their corresponding operators' repositories.
+
+* Good, because it centralizes the information about which versions are supported into each operator's repository
+* Good, because upgrades are predictable for the user, "upgrading the operator upgrades the cluster" is easy to explain and teach
+* Good, because image tags are stable and immutable once released
+* Bad, because we end up storing a lot of duplicate Docker images
+  * We could share the Docker layers to lessen this impact dramatically, but that would require rearchitecting our CI
```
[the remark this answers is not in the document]
teozkr: So I need to do some experiments here to confirm, but from what I remember this isn't quite true. Each layer also has a bunch of metadata including a UUID, which would be different unless the base layers were explicitly reused (such as by `docker build`'s cache mechanism).

> Adding a new tag to an image does not alter its layer digests.

Correct, but that would require the build process to have any knowledge of the prior build.

> If the operator is changed but not the product image then there is only an additional tag pointing to an already existing product image. No storage space is wasted.
> If the product image is changed but not the operator then the operator is nevertheless recompiled containing the new version number. If I used dive correctly to inspect the image of the Superset operator then the three image layers at the top are regenerated. In this case 20 MB are wasted.

Yes, that would be the ideal case.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: Nightlies are a good point to bring up... Personally, I'd rather tag nightlies with some form of build ID (commit hash?) than to start treating those image tags as mutable.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: Yes, or vice versa (product image must be released even if only the operator has any changes).

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+* Bad, because on-prem registry mirrors may not have whatever API or metadata we need to perform this resolution
+
+=== Configure `imagePullPolicy: Always` in deployed pods
+
+The operator always sets the `imagePullPolicy: Always` field in pods, forcing the Kubelets to always re-pull
+the image every time the Pod is started.
+
+* Good, because it's a minimal change
+* Bad, because it doesn't restart running pods, even if they are incompatible
+  * This means that it doesn't solve the original problem at all
```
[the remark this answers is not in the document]
teozkr: People were taking it seriously in the meeting (and you mention wanting to rely on the same mechanism elsewhere in this very review), so I'd rather leave it where it's more discoverable...

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,121 @@
...
+
+Currently, we publish all products as Docker images following the template
+`docker.stackable.tech/stackable/$PRODUCT:$PRODUCT_VERSION-stackable$IMAGE_VERSION`, as well as
+`docker.stackable.tech/stackable/$PRODUCT:$PRODUCT_VERSION-stackable$IMAGE_SEMVER_MAJOR`.
+
+In theory, a given Operator version depends on some SemVer range of `image_version` (for example `^1.3.0`, which
+expands to `>=1.3.0 <2.0.0`).
+
+In practice however, Docker tags are opaque, which prevents us from actually specifying this anywhere. Instead, we can only
+express the range `^N.0.0` (`>=N.0.0 <N+1.0.0`), depending on the `-stackable$IMAGE_SEMVER_MAJOR` tag. This works
```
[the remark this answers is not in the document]
teozkr: Tried to address this a bit more explicitly.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
teozkr: I'm not sure how much it'd help to centralize that cache, since each operator basically only cares about its own ~one image.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
soenkeliebau: ah

razvan:
I think it's worth looking at what others are doing in this regard. Google's Spark operator for example, provides a clear version matrix: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -0,0 +1,126 @@
...
+* Date: [YYYY-MM-DD when the decision was last updated] <!-- optional -->
+
+Technical Story: [description | ticket/issue URL] <!-- optional -->
+
+== Context and Problem Statement
+
+Currently, we publish all products as Docker images following the template
+`docker.stackable.tech/stackable/$PRODUCT:$PRODUCT_VERSION-stackable$IMAGE_VERSION`, as well as
+`docker.stackable.tech/stackable/$PRODUCT:$PRODUCT_VERSION-stackable$IMAGE_SEMVER_MAJOR`.
+
```
soenkeliebau, line 17: Note to @soenkeliebau : 
Mention requirements around whether or not operators should be able to support multiple product versions and able to install versions that were not released when the operator was released.

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -29,21 +47,131 @@ This isn't a breaking change for the product image, since it still works for old
...
+- 2.8.0-stackable0.1.0
+- 2.8.0-stackable0.2.0
+- 2.8.0-stackable0.1.1
+- 2.8.0-stackable0.1.2
+- 2.9.0-stackable0.1.2
+- 2.8.0-stackable0.2.1
+- 2.9.0-stackable0.2.1
+
+|*2* - Synchronize image versions in lockstep with operator versions - Variant 1
```
siegfriedweber, line 151:
```suggestion
|*2* - Synchronize image versions in lockstep with operator versions - Variant 2
```

modules/contributor/pages/adr/drafts/ADRx-product_image_versioning.adoc
```diff
@@ -29,21 +47,131 @@ This isn't a breaking change for the product image, since it still works for old
...
+- 2.9.0-stackable0.1.1
+
+|*3* - Use SemVer for docker images, independently of operator version - Variant 1
+a|
+
+- 2.8.0-stackable0.1.0
+- 2.8.0-stackable0.1.1
+- 2.9.0-stackable0.1.1
+
+|*4* - Use SemVer for docker images, independently of operator version - Variant 1
```
siegfriedweber, line 168:
```suggestion
|*4* - Use SemVer for docker images, independently of operator version - Variant 2
```

siegfriedweber (changes requested):

teozkr:
Looks reasonable to me. Might want to change title of one of them, and add the product image one to the nav toc.

modules/contributor/pages/adr/drafts/ADRx-operator_image_version_selection.adoc
```diff
@@ -0,0 +1,152 @@
+= ADRx: Product Image Versioning
```
siegfriedweber, line 1: The title of this document should be "ADRx: Operator image version selection". It currently contains only an outdated version of ADR18. So it would be better to just remove the entire file and restart from scratch when this topic comes up again.

There is only one noteworthy discussion we had in our meeting which should be written down somewhere: There must be a compatibility matrix for the operator and product versions. This matrix should probably contain at least the following states: "compatible" (which means tested and supported by Stackable), "unsupported" (which means there is no known technical restriction which prevents you from using it but it is either not tested or contains vulnerabilities), and "incompatible" (which means there were breaking changes and this combination will not work). This compatibility matrix should not be hard-coded into the operators because then a new release of an operator is required every time a new product version should be supported by this operator. Instead it should be read from a config map.

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
```diff
@@ -0,0 +1,252 @@
...
+* Old versions of operators must keep working and deploying images that they are compatible with
+* We should try to limit the number of maintained image version "tracks" for each product
+* It should not be necessary to update the used operator version in order to be able to deploy a new product version
+* The chosen option should make life as easy as possible for the end-user, it doesn't need to be the simplest one to implement
+
+== Considered Options
+There are two main options that were considered, both of which can be implemented in two variants:
+
+=== Option Variants
+==== Variant 1 - Git tags contains only the Stackable image version
```
siegfriedweber, line 76:
```suggestion
==== Variant 1 - Git tags contain only the Stackable image version
```

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
```diff
@@ -0,0 +1,252 @@
...
+|Git Tag |Docker Registry Tag
+
+|kafka2.8.0_stackable0.1.0
+|2.8.0_stackable0.1.0
+
+|kafka2.9.0_stackable0.1.0
+|2.9.0_stackable0.1.0
+
+|kafka2.9.0_stackable0.1.1
+|2.9.0_stackable0.1.2
```
siegfriedweber, line 195: This entry is not necessary. But if it is kept then the Stackable versions must not differ.

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
```diff
@@ -0,0 +1,252 @@
...
+|===
+
+A subsequent ADR will contain follow up decisions on how much of the selection process for the correct image we want to automate in the operators.
+Initially no automation will be implemented, users need to select a working product version and image version combination and refer to the fully qualified version from the CRD (i.e. docker registry tag from the table above).
+
+The main reasons for picking this option were:
+* it allows us to decouple operator version and image version to a high degree
```
siegfriedweber, line 202:
```suggestion
The main reasons for picking this option were:

* it allows us to decouple operator version and image version to a high degree
```

The new line is necessary. Otherwise the list is not rendered as a list.

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
```diff
@@ -0,0 +1,252 @@
...
+* there is no need to build _unneeded_ images just to accommodate changes in other components (operator change vs image change)
+* it keeps the option to provide automation around selecting the correct image version later on without the need for breaking changes
+
+NOTE: Image versions are only comparable when combined with the product version _0.1.0_ in kafka2.8.0_stackable0.1.0 and kafka2.9.0_stackable0.1.0 are two completely different versions _0.1.0_
```
siegfriedweber, line 206: This is hard to read. It should be split into two sentences.

siegfriedweber (changes requested):

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
soenkeliebau: The 2 was a typo, good catch!

But as for whether or not it is neccesary, I included it to highlight the fact that 0.1.1 can be released for one Kafka version but not the other.

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
soenkeliebau: done

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
soenkeliebau: done

modules/contributor/pages/adr/ADR018-product_image_versioning.adoc
soenkeliebau: done

siegfriedweber (approved):
LGTM

PR #155, opened 2022-02-25
Add secret-operator

teozkr:
Depends on https://github.com/stackabletech/secret-operator/pull/64

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -9,6 +9,10 @@ content:
     - url: https://github.com/stackabletech/documentation.git
       branches: main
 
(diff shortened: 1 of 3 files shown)


teozkr:
R4R now that https://github.com/stackabletech/secret-operator/issues/64 is merged.

maltesander (approved):
LGTM.

lfrancke:
This will also require adding it to the menu, right? I believe it's not there at the moment.

teozkr:
That's correct, since the menu is controlled by the UI bundle… :/

PR #157, opened 2022-03-01
Relativize home page links to operators

teozkr:


lfrancke (approved):
I haven't tested it but it looks reasonable

PR #158, opened 2022-03-01
Build Antora UI bundle from submodule rather than pulling from Nexus

teozkr:


.github/workflows/deploy.yml
```diff
@@ -13,12 +13,14 @@ jobs:
...
+        with:
+          submodules: true
       - uses: actions/setup-node@v2
         with:
           node-version: '16'
           cache: 'npm'
-      - run: npm install
-      - run: npx antora --fetch local-antora-playbook.yml
+      - run: npm ci
+      - run: make
```
lfrancke, line 23: Should we not include ANTORAFLAGS --fetch in the workflow?
In theory nothing bad should happen because we don't cache anything _at the moment_ but it's so easy to forget to add it should we ever start caching.

lfrancke (changes requested):

lfrancke (approved):

PR #164, opened 2022-03-15
Move navbar into documentation repo

teozkr:
This depends on https://github.com/stackabletech/documentation-ui/pull/8 for it to actually be respected.

### DIFF
--- a/supplemental-ui/js/search-ui.js
+++ b/supplemental-ui/js/search-ui.js
@@ -1,0 +1,1 @@
+../../node_modules/@antora/lunr-extension/supplemental_ui/js/search-ui.js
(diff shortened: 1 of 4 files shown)


maltesander (approved):
LGTM.

supplemental-ui/js/search-ui.js
```diff
@@ -0,0 +1 @@
+../../node_modules/@antora/lunr-extension/supplemental_ui/js/search-ui.js
```
lfrancke, line 1: I'm a bit confused. Is this an artifact of the Github PR preview thing or does the file really only contain this single line?

teozkr: It's a symlink to that file (which will be installed by npm).

teozkr: This is one of those annoying cases where it's called out by the individual file viewer:

image

and the file browser:

image

and even the diff browser file tree:

image

but for some reason it's completely ignored by the diff viewer itself:

image

lfrancke (approved):

PR #169, opened 2022-03-31
Add support for plantuml

sbernauer:
I'm not sure if https://github.com/Mogztter/asciidoctor-kroki is the right tool to use.
But I'm to stupid to install https://github.com/asciidoctor/asciidoctor-diagram
Anyone who knows this npm/ruby/gem/whatever stuff please feel free to tell me how to do it the right way

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -62,3 +62,6 @@ antora:
 asciidoc:
   attributes:
     base-repo: https://github.com/stackabletech
+    plantuml-server-url: http://www.plantuml.com/plantuml
+  extensions:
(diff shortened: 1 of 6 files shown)


lfrancke (approved):
I haven't tried it but from all I know it looks good!

EDIT: If you want me to actually test it it'll have to wait until tomorrow or so.

sbernauer:
I have tested the solution and it worked. The possible problem that i see is that it uses some online service (https://kroki.io/ i guess) and i don't know if we want a dependency on that.
I'm personally fine with it as our docs are non-sensitive and we can switch to something else if it causes any problems.

PR #178, opened 2022-04-13
Add draft ADR on configmap/secret content access for operators

soenkeliebau:


teozkr (dismissed):
Might want to add this to the nav.

Are we happy with the decision and ready to promote this to decided?

fhennig:
We talked about this in the Architecture meeting and everyone is happy with the decision, so it is decided

sbernauer (approved):
Thanks a lot guys!

PR #189, opened 2022-05-05
Fix names of ADR documents

sbernauer:
* Fixes description in (yet) ADR 17
* Fixes the hidden "ADR013: Supported Kubernetes versions" we searched yesterday
* Unifies file names
* Fixes typo in file names


fhennig (approved):
Thanks for this!

PR #193, opened 2022-05-09
Add a first version of the three-minute-quickstart page. 

soenkeliebau:
Mostly to give us a working link so that we can generate a QR code for the flyer.

### DIFF
--- a/modules/ROOT/pages/quickstart.adoc
+++ b/modules/ROOT/pages/quickstart.adoc
@@ -1,0 +1,23 @@
+= Quickstart
+
+This is the super-short getting started guide that should enable you to get something up and running in less than three minutes (excluding download times).
+
+== Setup
+
+Download the cluster script.
+
+----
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/quickstart.adoc
```diff
@@ -0,0 +1,23 @@
+= Quickstart
+
+This is the super-short getting started guide that should enable you to get something up and running in less than three minutes (excluding download times).
+
+== Setup
+
+Download the cluster script.
+
+----
+curl -L https://raw.githubusercontent.com/stackabletech/integration-tests/main/create_test_cluster.py > create_test_cluster.py
```
maltesander, line 10: Do we want to promote that script or point to stackablectl?

soenkeliebau: We don't want to promote that long term, we just need something for right now because there is a conference tomorrow.
This will be updated as soon as stackablectl reaches feature-parity.

Issue #187, opened 2022-05-03
Add documentation for common structs (S3, OPA)


fhennig:
We have common structs for:
- S3
- OPA

We need common documentation for these. This should probably be two-fold:
- A conceptual page: Why did we design it this way? How does it fit into the larger picture?
- A reference page: A page showing and describing all possible configuratoin settings

Guides about specific usage will be in the operators

PR #200, opened 2022-05-31
S3 resource documentation

fhennig:
This is part of ticket #187


modules/concepts/pages/s3.adoc
```diff
@@ -0,0 +1,214 @@
...
+            inline: # <2>
+              host: test-minio
+              port: 9000
+  # more spec here ...
+----
+<1> The inline definition of the bucket. The bucket definition contains `bucketName` and `connection`.
+<2> The inline definition of the connection. It contains the `host` and `port`.
+
+
+==== Dedicated resources
```
razvan, line 110: "Dedicated" doesn't sound like it separates them from inlined resources very well.  Maybe "free", "unwrapped" or "stand-alone" ?

modules/concepts/pages/s3.adoc
```diff
@@ -0,0 +1,214 @@
...
+  deepStorage:
+      s3:
+        reference: my-bucket-resource
+  # more spec here ...
+----
+
+== Credentials
+
+
+No matter if a connection is specified inline or as a separate object, the credentials are always specified in the same way. You will need a `Secret` containing the access key ID and secret access key, a `SecretClass` and then a reference to this `SecretClass` where you want to specify the credentials.
```
razvan, line 168: I feel an explanation is need  to "why" a SecretClass is necessary? Why can't I just reference the Secret directly ?

modules/concepts/pages/s3.adoc
fhennig: Good point, I've changed it to stand-alone

razvan (approved):
SecretClass clarifications will follow in a new PR

PR #203, opened 2022-06-01
Enabled tags

fhennig:
We have our first operator now that has a tagged version as well as a nightly version - the airflow operator.

This PR enables that the tagged version is also displayed, and it also enables this for all other operators.


razvan (approved):
lgtm

Issue #156, opened 2022-02-25
Versioned docs


lfrancke:
As a user of our software I'd like to be able to read documentation for the version of the platform that I'm running.

Currently we only ever publish the latest version of our documentation.
We do want to publish platform level documentation and later add all versions of all operators. For now we're happy with having the latest platform documentation available.

This is done when..
- [ ] There is a `latest` version of our documentation available that points to the current `main` branches of all our repositories
- [ ] https://docs.stackable.tech points to the latest _released_ version of our platform
- [ ] The documentation site has a way of selecting which version of the documentation to read
- [ ] There are documentation sites available for at least the last released platform version, older versions are optional

fhennig:
At the top of my head, I think this is not super trivial with Antora. Antora supports versions for components, and at the moment we have the "home" component for all platform stuff, and then each operator is a component too. We could easily use the antora version mechanism for the operator versions, but then how do we build the platform version on top?

lfrancke:
Thank you very much. I think all of that makes sense and I have nothing to add.
We can definitely get someone to help you out on the UI part.

fhennig:
This ticket is almost done. Update:
- We now publish documentation for our latest `main` state which is always `nightly` and also published under this name.
- We publish documentation from tags, the format `docs/X.Y` was introduced for these tags.
- We adapted the Makefile in operator-templating to automatically derive the antora component version from the cargo package version
- A dropdown to select a version shows up in the UI (top right) if multiple versions are present
- There is a `stable` link which always links to the latest stable version.

What's left to do:
- release a stable docs version for every component
- Adapt the menu links for all operators to point to the stable version (currently points to nightly)

PR #206, opened 2022-06-02
Enable 'stable' URLs

### LINKED ISSUE #198: "latest" redirect
Antora supports special treatment for the latest stable version of a component.

https://docs.antora.org/antora/latest/playbook/urls-latest-version-segment/

It is possible to have url with the segment `.../latest/...` which then replaces or redirects to the latest version (i.e. `.../0.4/...`)

for redirects, it depends on how the documentation is hosted (I'm not sure how we host our docs)

fhennig:
This is also related to #156 

This change enables a 'stable' url which points to the latest stable version of a documentation component. For example:
https://docs.stackable.tech/airflow/stable/index.html instead of https://docs.stackable.tech/airflow/0.3/index.html

closes #198

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -4,6 +4,11 @@ site:
   url: https://docs.stackable.tech
   robots: allow
 
+urls:
+  # This replaces the component version in the URL of the latest stable version with 'stable'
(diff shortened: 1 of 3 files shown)


razvan (approved):
works.

PR #219, opened 2022-06-21
Add instructions for using stackablectl instead of Helm

fhennig:
We want to promote the use of stackablectl. This PR adds the instructions for using it to the Getting started guide as well as the Tutorial. The Helm alternative is still provided.


Jimvin (approved):
LGTM

PR #220, opened 2022-06-22
Removed spark standalone everywhere

fhennig:
We are retiring the spark (standalone) operator (https://github.com/stackabletech/issues/issues/227) and part of that is removing the docs. This PR does that.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -60,10 +60,6 @@ content:
       branches: main
       tags: docs/*
       start_path: docs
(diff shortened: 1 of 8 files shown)


razvan:
I cannot comment on things that didn't change:

fhennig:
Hmm yes, that file has been a bit unmaintained, I'm not sure what it is for. I'll update it though.

razvan (approved):
beautiful

PR #221, opened 2022-06-23
Role and role group documentation

### LINKED ISSUE #191: Add documentation on concepts: Roles and Role Groups
We don't currently have documentation on the terms _role_ and _roleGroup_.

We should add a glossary or concepts section that explains these.

fhennig:
closes #191 

This adds a new concepts page to explain roles and role groups


fhennig:
it's not urgent

modules/concepts/pages/roles-and-role-groups.adoc
```diff
@@ -0,0 +1,80 @@
+= Roles and role groups
+
+Software - especially the distributed kind - often consists of multiple different process that work together to achieve their goal, Stackable calls these _roles_. The same kind of process is then often replicated multiple times. These replicas can further be subdivided into _role groups_. Roles and role groups are defined in the resource manifest.
```
maltesander, line 3:
```suggestion
Software - especially the distributed kind - often consists of multiple different processes that work together to achieve their goal. Think of a client communicating with a server. Stackable calls these _roles_. The roles or processes are often replicated multiple times. These replicas can further be subdivided into _role groups_. Roles and role groups are defined in the resource manifest.
```

modules/concepts/pages/roles-and-role-groups.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+== Motivation
+// configuration and scheduling
+The use of roles and role groups is motivated by the need to configure differently *how* processes are configured and *where* they run.
```
maltesander, line 7:
```suggestion
The use of roles and role groups is required to control *how* processes are configured and *where* they run.
```

modules/concepts/pages/roles-and-role-groups.adoc
```diff
@@ -0,0 +1,80 @@
...
+// node labels for location and hardware are typical uses
+// also nodes that might have GPU attached, or special storage
+
+While configuration is concerned with modifying how the process works on the inside, scheduling is concerned with the context where it should run inside the cluster. A processing task should be scheduled on a node with faster CPUs or nodes with GPUs attached. A caching process is best scheduled on a node with SSD storage, for example.
+
+This is done by using link:https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/[label selectors]. Nodes can be labeled and then processes can be assigned to nodes based on these labels.
+
+=== Role groups
+
+Role groups are motivated by the need to subdivide the replicas of a processes into groups as well, either to configure multiple groups differently or to schedule them differently. For example in a cluster spanning multiple physical locations, you may want to make sure half of your processes run in one location and half in the other.
```
maltesander, line 27:
```suggestion
Role groups are motivated by the need to subdivide the replicas of a process into groups as well, either to configure or to schedule multiple groups differently. For example in a cluster spanning multiple physical locations, you may want to make sure half of your processes run in one location and half in the other.
```

modules/concepts/pages/roles-and-role-groups.adoc
```diff
@@ -0,0 +1,80 @@
...
+
+== How it works
+
+A distributed software has multiple roles. Every role must be configured with at least one role group. Configuration can be specified at role level or at role group level, where the role group level overrides the role level. Label selectors for scheduling and the number of replicas are specified on the role group, not role. The role group is the lowest level, so it makes sense to have to specify this there.
```
maltesander, line 31:
```suggestion
A distributed software consists of one or more roles. Every role must be configured with at least one role group. Configuration can be specified at role level or at role group level, where the role group level overrides the role level. Label selectors for scheduling and the number of replicas are specified on the role group, not role. The role group is the lowest level, so it makes sense to have to specify this there.
```

modules/concepts/pages/roles-and-role-groups.adoc
```diff
@@ -0,0 +1,80 @@
...
+      hiCapacity:  # <4>
+        config:
+          resources:
+            storage:
+              data: 2Gi
+        replicas: 1
+  ...
+----
+
+<1> The JournalNode role with only a single default role group. For the role group 3 replicas are specified, specifying a replica count is not optional.
```
maltesander, line 73: This is not quite true. Replicas are optional and default to 1.
``` 
Replicas is the desired number of replicas of the given Template. These are replicas in the sense that they are instantiations of the same Template, but individual replicas also have a consistent identity. If unspecified, defaults to 1.
```

maltesander (changes requested):
One minor error mistakes and some comments. LGTM otherwise!

maltesander (approved):
LGTM.

PR #229, opened 2022-07-04
Minor fixes to example command

Jimvin:
The output from one of the example commands was left in an breaks the command if you use the copy/paste feature in the docs.


fhennig (approved):
LGTM

Issue #232, opened 2022-07-04
Callouts don't work with highlightjs


fhennig:
A [Callout](https://docs.asciidoctor.org/asciidoc/latest/verbatim/callouts/) is the little number in a circle to annotate source code listings. I've noticed that these don't work on source blocks with highlighting enabled, sadly. It seems like highlightjs is breaking them. I'm unsure what a fix could be.

Here's an example of what it looks like broken: https://docs.stackable.tech/druid/stable/usage.html#_s3_deep_storage

teozkr:
This seems to be related to our custom UI bundle, switching back to Antora's default bundle seems to "fix" it.

fhennig:
A bit of research results: 
- The breaking commit seems to be this one: https://github.com/stackabletech/documentation-ui/commit/d5c3d38c7ee9348ddb01ee552995384e5c8a781c Highlight.js was upgraded here, from v10 to v11 with some breaking changes.
- Callouts work in code blocks without highlighting, i.e. if only `[source]` instead of i.e. `[source,yaml]` is used.
- The default Antora UI has not yet upgraded to highlightjs 11, they are still at v9.

It seems to be that the new highlightjs is breaking the callouts.

fhennig:
We decided to downgrade to v10 and look at this again when Antora upgrades highlightjs in the Default UI

fhennig:
Hey, thanks for raising the issue there. I'm interested in seeing how it'll go!

PR #255, opened 2022-08-23
Update Antora to 3.1.0

### LINKED ISSUE #254: Upgrade to Antora 3.1
Antora 3.1 was released on the 22-08-2022. No major or breaking changes, an upgrade should be very easy.

release notes: https://docs.antora.org/antora/latest/whats-new/

teozkr:
Fixes #254 

Ran into this when getting annoyed about the broken docs callouts (#232). It doesn't seem to have fixed the issue, but it doesn't seem to have made things worse either...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,68 +12,69 @@
         "asciidoctor-kroki": "0.15.4"
       },
       "devDependencies": {
-        "@antora/cli": "3.0.1",
-        "@antora/site-generator": "3.0.1"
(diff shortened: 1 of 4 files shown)


fhennig (approved):
Why did you remove the search files? Were they not required?

Anyways I ran it locally and it looks like everything is fine.

teozkr:
Yeah the plugin now injects that itself.

PR #259, opened 2022-08-24
ADR on Product image selection

Maleware:
ADR22 on design of operators to accept custom images


modules/contributor/pages/adr/ADR022-image-selection-operator
```diff
@@ -0,0 +1,121 @@
+= ADR022: Image selection with operators
```
fhennig, line 1: the ADR number will be given to the ADR once it's decided, but just to let you know (I think the ADR process isn't really documented anywhere)

modules/contributor/pages/adr/ADR022-image-selection-operator
```diff
@@ -0,0 +1,121 @@
...
+** Malte Sander
+** Sebastian Bernauer
+** Felix Henning
+** Maximilian Wittich
+** Lars Franke (absent)
+** Soenke Liebau (absent)
+
+* Date: 22-08-24
+
+== Current State (example superset)
```
fhennig, line 16: The current state is that the product version + stackable version has to be specified, right? I'd omit some parts of the example CRD (everything that isn't relevant really).

Also part of this section should be: What is the problems that we have with the current state? I suspect the immediate thing that led to all this is that it's not possible to change the repo that the images are pulled from, right?

We could also say that it's not possible for a customer to mirror images or to provide their own images entirely, with different tags etc.

modules/contributor/pages/adr/ADR022-image-selection-operator
```diff
@@ -0,0 +1,121 @@
...
+
+apiVersion: superset.stackable.tech/v1alpha1
```
fhennig, line 18:
```suggestion
[source,yaml]
----
apiVersion: superset.stackable.tech/v1alpha1
```

start of code block with syntax highlighting

modules/contributor/pages/adr/ADR022-image-selection-operator
```diff
@@ -0,0 +1,121 @@
...
+  name: superset
+spec:
+  version: 1.4.1-stackable2.1.0
+  statsdExporterVersion: v0.22.4
+  credentialsSecret: superset-credentials
+  loadExamplesOnInit: false
+  nodes:
+    roleGroups:
+      default:
+        replicas: 1
```
fhennig, line 30:
```suggestion
        replicas: 1
----
```

end of code block

modules/contributor/pages/adr/ADR022-image-selection-operator
```diff
@@ -0,0 +1,121 @@
...
+** users specified version 1.4.1-stackable2.1.0 and imageRepository my.corp/superset
+=> operator should use image my.corp/superset:1.4.1-stackable2.1.0 based on imageRepository
+=> Equivalent to user saying use the image "my.corp/superset:1.4.1-stackable2.1.0"
+** users specified version 1.4.1 and imageRepository my.corp/superset
+=> operator should use image my.corp/superset:1.4.1-stackable2.1.0 based on imageRepository and compability knowledge (which operator works with which image)
+=> Equivalent to user saying use the image "my.corp/superset:1.4.1-stackable2.1.0"
+=> NOT equivalent to user saying use the image "my.corp/superset:1.4.1"
+=> When specifing the image directly users loose the ability to let the operator determine its compatible stackable image version
+
+== Decision Drivers
```
fhennig, line 107: I'd put the decision drivers before the list of options, so I have those in mind when reading through the options.

I also had additional points in my notes from the arch meeting:
- allow the user to easily paste in a complete image name (suggested by Teo, convenient for testing)
- allow users to use some defaults, allow them to only override certain things (suggested by Sigi, i.e. only override the repo or only override the version etc., but I think that's already there in the list of decision drivers)

modules/contributor/pages/adr/ADR022-image-selection-operator
```diff
@@ -0,0 +1,121 @@
...
+
+* One time work to create a flexible and extendable solution for possible use cases
+* Preserve the possibility of magic version completition of operators for images (e.g. superset:1.4.1 => superset:1.4.1-stackable2.1.0) 
+
+== Decision Outcome
+
+Antendees made the decision for Option 5, since it covers almost all use cases except that with costum repositories the autocompletion magic will not work.
+We think, that this is a corner case because the probability of a user mirroring our whole image repository (docker.stackable.tech/stackable) to a custom repository
+and still want the magic mechanism to work seems to be small.
+Furthermore, Option 5 will provide enough flexibility to effortless build that specific usecase if really needed.
```
fhennig, line 117: Maybe mention complex enum

fhennig (commented):
Thanks for doing this Maxi!

I added some notes. Also it looks like the file doesn't have the `.adoc` extension (which also causes it to not be rendered by github)

modules/contributor/pages/adr/ADR022-image-selection-operator
Maleware: Okay, I just remove the number then.

modules/contributor/pages/adr/ADR022-image-selection-operator
Maleware: Okay, it's done :)

modules/contributor/pages/adr/ADR022-image-selection-operator
Maleware: Done

modules/contributor/pages/adr/ADR022-image-selection-operator
Maleware: Done

modules/contributor/pages/adr/ADR022-image-selection-operator
Maleware: Yeah, that's a great Idea. Thank you!

modules/contributor/pages/adr/ADR022-image-selection-operator
Maleware: I did :)

fhennig:
I want to link this ticket: https://github.com/stackabletech/issues/issues/141 as it is relevant

maltesander:
I did a quick test with an untagged enum https://github.com/stackabletech/operator-rs/blob/image_version_crd/src/commons/image.rs (dont mind the names, i did not use the ADR names yet). 
This leads to problems. First of the `Custom` and `Auto` variants get confused when deserializing, which causes the tests to fail. Secondly i cannot apply a CRD using this as all:
```
[customresourcedefinition.apiextensions.k8s.io/zookeeperznodes.zookeeper.stackable.tech](http://customresourcedefinition.apiextensions.k8s.io/zookeeperznodes.zookeeper.stackable.tech) created
The CustomResourceDefinition "zookeeperclusters.zookeeper.stackable.tech" is invalid:
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[0].properties[productVersion].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[0].properties[repo].default: Forbidden: must be undefined to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[0].properties[repo].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[0].properties[stackableVersion].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[0].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[1].properties[custom].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[1].properties[productVersion].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].anyOf[1].type: Forbidden: must be empty to be structural
* spec.validation.openAPIV3Schema.properties[spec].properties[version].type: Required value: must not be empty for specified object fields
```
@teozkr said this could probably be fixed in kube-rs. But until then, our proposed solution does not work unless we actually use the enum variant names in the CRD (not using untagged), which will be another breaking change when trying to move back to the untagged version.

sbernauer:
Thanks @maltesander!
I feared the same thing, so we have to use a complex enum I'd say

sbernauer:
We also want a new attribute called `pullSecrets` similar to `imagePullSecrets` in spark-k8s setting https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret
```
  pullPolicy: IfNotPresent # String. defaults to IfNotPresent
  pullSecrets:
    - name: regcred # reference to secret in same namespace
```

modules/contributor/pages/adr/ADR023-image-selection-operator.adoc
```diff
@@ -0,0 +1,186 @@
...
+
+* Bad, because it's breaking
+* Good, because it gives all flexibility of all previous options
+* Good, because we can non-breaking introduce new "magic" in the future by adding new image enum variants
+* Good, because we can implement it as enum called e.g. `ImageSpec` in operator-rs which will offer a function like `resolve_image` that will make it easy for operators to use
+
+[[option6]]
+=== Option 6
+
+This option is **breaking**. It uses a complex enum, similar to Option 5. Option 5 does not account for the need to specify the product version with a custom image. It is also not possible to just use a custom docker repository and still use the operator recommended version (i.e. just mirroring the stackable repository). This option makes that possible. We first start with implementing stackableVersion and custom. The stackable enum will be implemented as soon as we have magic versioning.
```
sbernauer, line 128:
```suggestion
This option is **breaking**. It uses a complex enum, similar to Option 5. Option 5 does not account for the need to specify the product version with a custom image. It is also not possible to just use a custom docker repository and still use the operator recommended version (i.e. just mirroring the stackable repository). This option makes that possible. We first start with implementing `stackableVersion` and `custom`. The `stackable` enum variant will be implemented as soon as we have the magic stackableVersion resolution. 
```

modules/contributor/pages/adr/ADR023-image-selection-operator.adoc
```diff
@@ -0,0 +1,186 @@
+= ADR023: Product image selection
+Maxi Wittich <maximilian.wittich@stackable.tech>, Sebastian Bernauer <sebastian.bernauer@stackable.tech>
+v0.1, 2022-08-31
+:status: pending
```
sbernauer, line 4:
```suggestion
:status: accepted
```

sbernauer (approved):
LGTM % commets.
Can you please rename the file to something like `ADR023-product-image-selection.adoc` (don't forget to update in current_adrs.adoc)

Maleware:
@sbernauer done

sbernauer (approved):

Maleware:
bors merge

PR #266, opened 2022-09-01
Add versioning to the main documentation

### LINKED ISSUE #240: Start versioning the platform docs as well
At the moment the latest stable and current nightly docs for the platform docs have not diverged and it is okay to not have versions, but it should be versioned in the long term.

Like with the operators, it needs to have tags and versions at those tags. But there is no release script, so it can be adapted manually.

- [ ] for the current docs, the version `22.06` needs to be put into the `version` key in the `antora.yaml`. This needs to be committed and tagged with `docs/22.06`
- [ ] afterwards, another commit should set the `version` key to `nightly` and `prerelease: true`
- [ ] in the `antora-playbook.yaml` the `tags: docs/*` needs to be added to the documentation repository, so the tagged versions are pulled in

Open question: Should there be an automated mechanism to release new versions? To release a new version, the `version` and `prerelease` keys in `antora.yaml` need to be adapted and the commit needs to be tagged. then a new commit needs to be made to set the version back to nightly and the prerelease to true. (This could be put into a script, possibly in the makefile?)

fhennig:
This fixes #240 which we will need for the next release. This PR

- Starts versioning immediately with a 22.06 and a nightly version
- sets up tag based versioning for the future
- adds a script to simplify the version bumping in the future
- Adapts the menu links (they broke when versioning was introduced)


maltesander (approved):
LGTM.

Issue #233, opened 2022-07-04
Upgrade documentation 2022-06 -> 2022-09


lfrancke:
Needs to be done by the time of the release
Copied from https://github.com/stackabletech/.github-private/blob/main/profile/README.md:

- [x] Document how to use stackablectl to uninstall all and install new release
- [x] Document how to use helm to uninstall all and install new release
- [x] Every breaking change of all our operators
- [x] List dropped supported product versions (if there are some)
- [x] List dropped supported operators (if there are some)
- [x] List supported k8s versions

adwk67:
> Putting this on track as I'd like to wait with merging for the actual release

_Originally posted by @sbernauer in https://github.com/stackabletech/documentation/issues/268#issuecomment-1236906140_

lfrancke:
We need to make sure that https://github.com/stackabletech/trino-operator/pull/278 is covered (I haven't looked at the PR, it might already be)

sbernauer:
Docs are adopted to to reflect trino-op 0.6.0 :heavy_check_mark:

PR #268, opened 2022-09-05
Document upgrade 22.06 -> 22.09

sbernauer:
Closes https://github.com/stackabletech/documentation/issues/233

### DIFF
--- a/modules/ROOT/pages/release_notes.adoc
+++ b/modules/ROOT/pages/release_notes.adoc
@@ -12,15 +12,13 @@ The main features focus on OpenShift support and security.
 The following new major platform features were added:
 
 OpenShift compatibility::
-As a first step towards certification of our own SDP operators we are implementing stricter security rules so that compatibility is satisfied.
-This work will be iterative in nature but similar security features - mainly the fact that product Pods must run with a custom `ServiceAccount` and `SecurityContextConstraint` - will be added to other operators over time.
-With this release the following operators have been improved with the previously mentioned features:
+We have made continued progress towards OpenShift compability, and the following operators can now be previewed on OpenShift.
+Further improvements are expected in future releases, but no stability or compatibility guarantees are currently made for OpenShift clusters.
 
 * https://github.com/stackabletech/airflow-operator/pull/127[Apache Airflow]
 * https://github.com/stackabletech/hbase-operator/pull/232[Apache HBase]
 * https://github.com/stackabletech/hdfs-operator/pull/225[Apache HDFS]
 * https://github.com/stackabletech/spark-k8s-operator/pull/126[Apache Spark on K8s]
-
 
 Support for internal and external TLS::
 The following operators support operating the products at a maximal level of transport security by using TLS certificates to secure internal and external communication:
--- a/modules/ROOT/pages/release_notes.adoc
+++ b/modules/ROOT/pages/release_notes.adoc
@@ -3,6 +3,205 @@
 The Stackable platform consists of multiple operators that work together.
 Periodically a platform release is made, including all components of the platform at a specific version.
 
+== Release 22.09
+This is our second release of the Stackable Data Platform.
+It contains lots of new features and bugfixes.
+The main features focus on OpenShift support and security.
+
+=== New platform features
+The following new major platform features were added:
+
+OpenShift compatibility::
+The following operators are now OpenShift compatible:
+
+* https://github.com/stackabletech/airflow-operator/pull/127[airflow]
+* https://github.com/stackabletech/hbase-operator/pull/232[hbase]
+* https://github.com/stackabletech/spark-k8s-operator/pull/126[spark-k8s]
+
+Support for internal and external TLS::
+The following operators support operating the products at a maximal level of transport security by using TLS certificates to secure internal and external communication:
+
+* https://github.com/stackabletech/trino-operator/pull/244[trino]
+* https://github.com/stackabletech/zookeeper-operator/pull/479[zookeeper]
+
+LDAP authentication::
+Use a central LDAP server to manage all of your user identities in a single place.
+The following operators added support for LDAP authentication:
+
+* https://github.com/stackabletech/airflow-operator/pull/133[airflow]
+* https://github.com/stackabletech/nifi-operator/pull/303[nifi]
+* https://github.com/stackabletech/superset-operator/pull/180[superset]
+
+=== stackablectl
+
+`stackablectl` now supports deploying ready-to-use demos, which give and end-to-end demonstration of the usage of the Stackable data platform.
+The xref:stackablectl::quickstart.adoc[quickstart guide] shows how to get started with `stackablectl`. Here you can see the xref:stackablectl::demos/index.adoc[available demos].
+
+=== Supported Kubernetes versions
+This release supports the following Kubernetes versions:
+
+* `1.24`
+* `1.23`
+* `1.22`
+
+Support for `1.21` was dropped.
+
+=== Upgrade from 22.06
+==== Using stackablectl
+You can list the available releases as follows
+
+[source,console]
+----
+$ stackablectl release list
+RELEASE            RELEASE DATE   DESCRIPTION
+22.06              2022-06-30     First official release of the Stackable Data Platform
+22.09              2022-09-09     Second release focusing on OpenShift support and security
+----
+
+To uninstall the `22.06` release run
+
+[source,console]
+----
+$ stackablectl release uninstall 22.06
+[INFO ] Uninstalling release 22.06
+[INFO ] Uninstalling airflow operator
+[INFO ] Uninstalling commons operator
+# ...
+----
+
+To install the `22.09` release run
+
+[source,console]
+----
+$ stackablectl release install 22.09
+[INFO ] Installing release 22.09
+[INFO ] Installing airflow operator in version 0.5.0
+[INFO ] Installing commons operator in version 0.3.0
+[INFO ] Installing druid operator in version 0.7.0
+[INFO ] Installing hbase operator in version 0.4.0
+[INFO ] Installing hdfs operator in version 0.5.0
+[INFO ] Installing hive operator in version 0.7.0
+[INFO ] Installing kafka operator in version 0.7.0
+[INFO ] Installing nifi operator in version 0.7.0
+[INFO ] Installing opa operator in version 0.10.0
+[INFO ] Installing secret operator in version 0.6.0
+[INFO ] Installing spark-k8s operator in version 0.5.0
+[INFO ] Installing superset operator in version 0.6.0
+[INFO ] Installing trino operator in version 0.5.0
+[INFO ] Installing zookeeper operator in version 0.11.0
+# ...
+----
+
+==== Using helm
+Use `helm list` to list the currently installed operators.
+
+You can use the following command to install all of the operators that are part of the release 22.06:
+
+[source,console]
+----
+$ helm uninstall airflow-operator commons-operator druid-operator hbase-operator hdfs-operator hive-operator kafka-operator nifi-operator opa-operator secret-operator spark-k8s-operator superset-operator trino-operator zookeeper-operator
+release "airflow-operator" uninstalled
+release "commons-operator" uninstalled
+# ...
+----
+
+To install the release 22.09 run
+
+[source,console]
+----
+$ helm repo add stackable https://repo.stackable.tech/repository/helm-stable/
+$ helm install --wait airflow-operator stackable/airflow-operator --version 0.5.0
+$ helm install --wait commons-operator stackable/commons-operator --version 0.3.0
+$ helm install --wait druid-operator stackable/druid-operator --version 0.7.0
+$ helm install --wait hbase-operator stackable/hbase-operator --version 0.4.0
+$ helm install --wait hdfs-operator stackable/hdfs-operator --version 0.5.0
+$ helm install --wait hive-operator stackable/hive-operator --version 0.6.0
+$ helm install --wait kafka-operator stackable/kafka-operator --version 0.6.0
+$ helm install --wait nifi-operator stackable/nifi-operator --version 0.6.0
+$ helm install --wait opa-operator stackable/opa-operator --version 0.10.0
+$ helm install --wait secret-operator stackable/secret-operator --version 0.6.0
+$ helm install --wait spark-k8s-operator stackable/spark-k8s-operator --version 0.5.0
+$ helm install --wait superset-operator stackable/superset-operator --version 0.6.0
+$ helm install --wait trino-operator stackable/trino-operator --version 0.5.0
+$ helm install --wait zookeeper-operator stackable/zookeeper-operator --version 0.11.0
+----
+
+==== Breaking changes
+
+You will need to adapt your existing CRDs to the following breaking changes:
+
+==== druid-operator
+1. HDFS deep storage is now configurable via HDFS discovery config map instead of an url to a HDFS name node (https://github.com/stackabletech/druid-operator/pull/262[#262]).
+Instead of
+
+[source,yaml]
+----
+  deepStorage:
+    hdfs:
+      storageDirectory: hdfs://druid-hdfs-namenode-default-0:8020/data
+----
+
+use
+
+[source,yaml]
+----
+  deepStorage:
+    hdfs:
+      configMapName: druid-hdfs
+      directory: /druid
+----
+
+==== trino-operator
+1. TrinoCatalogs now have their own CRD object and get referenced by the TrinoCluster (https://github.com/stackabletech/trino-operator/pull/263[#263]). Instead of
+
+[source,yaml]
+----
+apiVersion: trino.stackable.tech/v1alpha1
+kind: TrinoCluster
+# ...
+spec:
+  hiveConfigMapName: hive
+  s3:
+    inline:
+      host: minio
+      port: 9000
+      accessStyle: Path
+      credentials:
+        secretClass: s3-credentials
+  # ...
+----
+
+use
+
+[source,yaml]
+----
+apiVersion: trino.stackable.tech/v1alpha1
+kind: TrinoCluster
+# ...
+spec:
+  catalogLabelSelector:
+    trino: simple-trino
+  # ...
+---
+apiVersion: trino.stackable.tech/v1alpha1
+kind: TrinoCatalog
+metadata:
+  name: hive
+  labels:
+    trino: trino
+spec:
+  connector:
+    hive:
+      metastore:
+        configMap: hive
+      s3:
+        inline:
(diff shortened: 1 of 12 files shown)


modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,205 @@
...
+This release supports the following Kubernetes versions:
+
+* `1.24`
+* `1.23`
+* `1.22`
+
+Support for `1.21` was dropped.
+
+=== Upgrade from 22.06
+==== Using stackablectl
```
adwk67, line 50: Can we add a note in the Release [PR](https://github.com/stackabletech/issues/issues/247) to make sure we test uninstall/install, once the operators are released?

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,205 @@
...
+
+[source,console]
+----
+$ helm repo add stackable https://repo.stackable.tech/repository/helm-stable/
+$ helm install --wait airflow-operator stackable/airflow-operator --version 0.5.0
+$ helm install --wait commons-operator stackable/commons-operator --version 0.3.0
+$ helm install --wait druid-operator stackable/druid-operator --version 0.7.0
+$ helm install --wait hbase-operator stackable/hbase-operator --version 0.4.0
+$ helm install --wait hdfs-operator stackable/hdfs-operator --version 0.5.0
+$ helm install --wait hive-operator stackable/hive-operator --version 0.6.0
```
adwk67, line 118: We have 0.7.0 for stackablectl install:
```suggestion
$ helm install --wait hive-operator stackable/hive-operator --version 0.7.0
```

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,205 @@
...
+$ helm install --wait kafka-operator stackable/kafka-operator --version 0.6.0
```
adwk67, line 119:
```suggestion
$ helm install --wait kafka-operator stackable/kafka-operator --version 0.7.0
```

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,205 @@
...
+$ helm install --wait nifi-operator stackable/nifi-operator --version 0.6.0
```
adwk67, line 120:
```suggestion
$ helm install --wait nifi-operator stackable/nifi-operator --version 0.7.0
```

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,205 @@
...
+$ helm install --wait trino-operator stackable/trino-operator --version 0.5.0
+$ helm install --wait zookeeper-operator stackable/zookeeper-operator --version 0.11.0
+----
+
+==== Breaking changes
+
+You will need to adapt your existing CRDs to the following breaking changes:
+
+==== druid-operator
+1. HDFS deep storage is now configurable via HDFS discovery config map instead of an url to a HDFS name node (https://github.com/stackabletech/druid-operator/pull/262[#262]).
```
adwk67, line 134:
```suggestion
1. HDFS deep storage is now configurable via the HDFS discovery config map instead of a url to a HDFS name node (https://github.com/stackabletech/druid-operator/pull/262[#262]).
```

adwk67 (commented):
Just some minor changes.

modules/ROOT/pages/release_notes.adoc
sbernauer: The outputs in this documents are fictive and not able to run yet. I would consider running them at the check-point "Update stackablectl/release.yaml to add the new release".
I'd say whoever adds the release to release.yaml will always give it a try

adwk67: Fine by me.

adwk67 (dismissed):

sbernauer:
Many thanks!

sbernauer:
Putting this on track as I'd like to wait with merging for the actual release

sbernauer:
Currently waiting on https://github.com/stackabletech/release/pull/11 to get the final release description

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,205 @@
...
+
+=== New platform features
+The following new major platform features were added:
+
+OpenShift compatibility::
+The following operators are now OpenShift compatible:
+
+* https://github.com/stackabletech/airflow-operator/pull/127[airflow]
+* https://github.com/stackabletech/hbase-operator/pull/232[hbase]
+* https://github.com/stackabletech/spark-k8s-operator/pull/126[spark-k8s]
```
adwk67, line 19: I think we have HDFS, too:
```suggestion
* https://github.com/stackabletech/hdfs-operator/pull/225[hdfs]
* https://github.com/stackabletech/spark-k8s-operator/pull/126[spark-k8s]

```

sbernauer: thx!

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,207 @@
...
+* https://github.com/stackabletech/airflow-operator/pull/127[airflow]
+* https://github.com/stackabletech/hbase-operator/pull/232[hbase]
+* https://github.com/stackabletech/hdfs-operator/pull/225[hdfs]
+* https://github.com/stackabletech/spark-k8s-operator/pull/126[spark-k8s]
+
+
+Support for internal and external TLS::
+The following operators support operating the products at a maximal level of transport security by using TLS certificates to secure internal and external communication:
+
+* https://github.com/stackabletech/trino-operator/pull/244[trino]
```
adwk67, line 26: Adding Kafka following the update to the kafka changelog:
```suggestion
* https://github.com/stackabletech/kafka-operator/pull/466[kafka]
* https://github.com/stackabletech/trino-operator/pull/244[trino]
```

sbernauer: Sure thing, thx!

sbernauer: I would rather us https://github.com/stackabletech/kafka-operator/pull/442 if that's ok?

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,208 @@
...
+It contains lots of new features and bugfixes.
+The main features focus on OpenShift support and security.
+
+=== New platform features
+The following new major platform features were added:
+
+OpenShift compatibility::
+The following operators are now OpenShift compatible:
+
+* https://github.com/stackabletech/airflow-operator/pull/127[airflow]
```
soenkeliebau, line 17: Nit: can we pretty up the link text a bit?

- Apache Airflow
- Apache HBase
- Apache HDFS
- Apache Spark on K8s

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,208 @@
...
+
+* https://github.com/stackabletech/trino-operator/pull/244[trino]
+* https://github.com/stackabletech/kafka-operator/pull/442[kafka]
+* https://github.com/stackabletech/zookeeper-operator/pull/479[zookeeper]
+
+LDAP authentication::
+Use a central LDAP server to manage all of your user identities in a single place.
+The following operators added support for LDAP authentication:
+
+* https://github.com/stackabletech/airflow-operator/pull/133[airflow]
```
soenkeliebau, line 34: See comment above about prettier link names.

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,208 @@
...
+[source,console]
+----
+$ stackablectl release uninstall 22.06
+[INFO ] Uninstalling release 22.06
+[INFO ] Uninstalling airflow operator
+[INFO ] Uninstalling commons operator
+# ...
+----
+
+To install the `22.09` release run
```
soenkeliebau, line 75: Does this work, or would this leave the old, now incompatible CRD (the actual definitions, not the objects) in place? 
Wasn't that an issue with Helm uninstalling stuff, that it doesn't touch the crds and then the next install doesn't overwrite/merge them?

soenkeliebau: If this is an issue, it is probably caused by us not versioning our CRDs properly, which we should try and fix soon-ish I guess :)
I've created https://github.com/stackabletech/documentation/issues/273 for this to at least get us started.

sbernauer: I added a command to remove the crds. What do you think about that?

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,216 @@
...
+* https://github.com/stackabletech/trino-operator/pull/244[Trino]
+* https://github.com/stackabletech/kafka-operator/pull/442[Apache Kafka]
+* https://github.com/stackabletech/zookeeper-operator/pull/479[zookeeper]
+
+LDAP authentication::
+Use a central LDAP server to manage all of your user identities in a single place.
+The following operators added support for LDAP authentication:
+
+* https://github.com/stackabletech/airflow-operator/pull/133[Apache Airflow]
+* https://github.com/stackabletech/nifi-operator/pull/303[Apache NifFi]
```
soenkeliebau, line 35:
```suggestion
* https://github.com/stackabletech/nifi-operator/pull/303[Apache NiFi]
```

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,216 @@
...
+* https://github.com/stackabletech/hdfs-operator/pull/225[Apache HDFS]
+* https://github.com/stackabletech/spark-k8s-operator/pull/126[Apache Spark on K8s]
+
+
+Support for internal and external TLS::
+The following operators support operating the products at a maximal level of transport security by using TLS certificates to secure internal and external communication:
+
+* https://github.com/stackabletech/trino-operator/pull/244[Trino]
+* https://github.com/stackabletech/kafka-operator/pull/442[Apache Kafka]
+* https://github.com/stackabletech/zookeeper-operator/pull/479[zookeeper]
```
soenkeliebau, line 28:
```suggestion
* https://github.com/stackabletech/zookeeper-operator/pull/479[Apache ZooKeeper]
```

modules/ROOT/pages/release_notes.adoc
soenkeliebau: Sounds good. As long as the order of operations is followed this shouldn't break anything. If someone runs this command with the operators still running, could it cause them to destroy all clusters? Probably not, because they'd just error out when trying to list a non-existing resource now that I think of it ...

sbernauer: I tend to agree.
@teozkr you posted `you should be able to reapply the CRD YAMLs without deleting the old ones`. Can you please elaborate on that, is there something we can do here?

teozkr: > Sounds good. As long as the order of operations is followed this shouldn't break anything. If someone runs this command with the operators still running, could it cause them to destroy all clusters? Probably not, because they'd just error out when trying to list a non-existing resource now that I think of it ...

Deleting the CRDs deleted all CRs of that kind. This has nothing to do with CRD versioning (which is about maintaining compatibility).

> @teozkr you posted you should be able to reapply the CRD YAMLs without deleting the old ones. Can you please elaborate on that, is there something we can do here?

`kubectl apply -f path-to-updated-crds.yaml` (or the equivalent via the API)

Do not delete the old CRDs. Do not delete anything. Do not pass go. Do not collect $200.

sbernauer: In the case of stackablectl this currently calls helm uninstall operator --version old && helm install operator --version new.
Does the helm install of the new version achieve the wanted behavior?

sbernauer: > No, Helm won't touch the CRD if it already exists in the cluster.

In that case if fear for now we have to add the command to delete the CRDs.

> Also, why are you uninstalling and reinstalling rather than using helm upgrade?

Good point. Currently stackablectl only support uninstall and install of releases. Never used `helm upgrade` to be honest

teozkr: > In that case if fear for now we have to add the command to delete the CRDs.

What's blocking us from deploying the updated CRDs?

sbernauer: The missing functionality in stackablectl. I mean we can also just offer 12 `helm upgrade` commands and don't use stackablectl

sbernauer: FYI: helm uninstall && helm install && ka https://raw.githubusercontent.com/stackabletech/trino-operator/main/deploy/helm/trino-operator/crds/crds.yaml produces
```
│ 2022-09-12T06:37:52.622390Z ERROR stackable_operator::logging::controller: Failed to reconcile object controller.name="trinoclusters.trino.stackable.tech" error=event queue error error.sources=[error returned by apiserver during wat │
│ ch: too old resource version: 3708 (3848): Expired, too old resource version: 3708 (3848): Expired] 
```

teozkr: That's fairly normal but the operator should recover from that automatically.

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,267 @@
...
+[INFO ] Uninstalling commons operator
+# ...
+----
+
+Afterwards you will need to update the CustomResourceDefinitions (CRDs) installed by the Stackable Platform.
+The reason is, that helm will uninstall the operators but not the CRDs.
+
+[source,console]
+----
+$ kubectl apply -f https://raw.githubusercontent.com/stackabletech/airflow-operator/0.5.0/deploy/helm/airflow-operator/crds/crds.yaml
```
teozkr, line 80: FWIW these can also be merged into one long command line... Maybe it would make sense to provide a YAML installing all 22.09 CRDs in one place?

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,267 @@
...
+RELEASE            RELEASE DATE   DESCRIPTION
+22.09              2022-09-09     Second release focusing on security and OpenShift support
+22.06              2022-06-30     First official release of the Stackable Data Platform
+----
+
+To uninstall the `22.06` release run
+
+[source,console]
+----
+$ stackablectl release uninstall 22.06
```
teozkr, line 68: I guess this is more of a question about stackablectl than about these release docs, but... Why on earth do you need to provide the version number to _uninstall_? What happens if it doesn't match the release you installed previously?

sbernauer: Reason is that different versions of the platform can ship with different operators. We already removed the spark operator and may add jupyterhub in the future.

modules/ROOT/pages/release_notes.adoc
sbernauer: The [source,console] - at least for me - has the ability to copy the commands, and produces a single command when copying. I personally like that behavior because readability and ease of use

Regrading a release-crds yaml: Yeah we could do this, but on the other hand, the question is, if it is worth the effort by now

modules/ROOT/pages/release_notes.adoc
```diff
@@ -3,6 +3,267 @@
...
+== Release 22.09
+This is our second release of the Stackable Data Platform.
+It contains lots of new features and bugfixes.
+The main features focus on OpenShift support and security.
+
+=== New platform features
+The following new major platform features were added:
+
+OpenShift compatibility::
+The following operators are now OpenShift compatible:
```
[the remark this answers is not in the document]
sbernauer: @razvan could you please help a little with the correct wording? I have to admit they I'm not entirely sure what we can safely claim here

modules/ROOT/pages/release_notes.adoc
teozkr: Yeah but you could do something like:

```
kubectl apply \
  -f https://raw.githubusercontent.com/stackabletech/airflow-operator/0.5.0/deploy/helm/airflow-operator/crds/crds.yaml \
  -f https://raw.githubusercontent.com/stackabletech/commons-operator/0.3.0/deploy/helm/commons-operator/crds/crds.yaml \
...
```

Which should still be copy/pasteable. But meh, it works as is.

modules/ROOT/pages/release_notes.adoc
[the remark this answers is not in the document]
razvan: Yeah, there are a lot of _"buts"_ to be mentioned here.

modules/ROOT/pages/release_notes.adoc
teozkr: Isn't that our responsibility to store somewhere then? But yeah, too late for 22.09…

sbernauer: Absolutely fine to change that behavior, it#s far from perfect

modules/ROOT/pages/release_notes.adoc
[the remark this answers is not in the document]
sbernauer: Ah many thx!
I than suggest something like

As a first step towards certification of our own SDP operators we are implementing stricter security rules so that compatibility is satisfied. This work will be iterative in nature but similar security features - mainly the fact that product Pods must run with a custom `ServiceAccount` and `SecurityContextConstraint` - will be added to other operators over time. With this release the following operators have been improved with the previously mentioned features:

modules/ROOT/pages/release_notes.adoc
teozkr: Created an issue for this: https://github.com/stackabletech/stackablectl/issues/117

sbernauer:
Hi, could you please take another look so we can get this sorted out today?

modules/ROOT/pages/release_notes.adoc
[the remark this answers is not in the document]
teozkr: I would weaken it to a generic statement about making progress towards compatibility. Labelling anything as "compatible" is a pretty vague claim that can induce some assumptions that I'm personally not comfortable with at the moment:

1. Installation via OLM (even if not certified yet)
   - This requires us to figure out how to package the CSI operators (secret, listener), and solve https://bugzilla.redhat.com/show_bug.cgi?id=2100429
2. Regular regression testing
3. A commitment to future compatibility
   - While we do want to support OpenShift, we're still not sure about how to handle secret-op and listener-op. Committing to OS compatibility before we have that figured out would prevent us from integrating those into SDP in the meantime.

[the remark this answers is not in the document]
sbernauer: many thanks!

teozkr (approved):

sbernauer:
Thanks all of you!

PR #274, opened 2022-09-09
Adapt GitHub workflows to produce versioned platform documentation

### LINKED ISSUE #269: Bug: Versioned platform docs do not work
Versioned platform docs do work locally, but not online. 

The github actions (build, deploy) uses the `local-antora-playbook.yml` to build the docs for the website. This means it builds from the local github repo inside of the action. The checkout step in the action only fetches the latest commit, so the tags/commits are missing. Using a local path instead of a URL also activates [author mode](https://docs.antora.org/antora/latest/playbook/author-mode), but it seems to be without consequences in our case.

Solution:
- **deploy action** - For the deploy action I'd use `make PLAYBOOK=antora-playbook.yml` to use the other playbook. This other playbook specifies a URL for the documentation repository, so documentation will be pulled from there. This is a simple change in the action yaml file.
- **build action** - The build action is used in every PR to check if the docs build correctly. In there its important to build from the local branch, so using the `local-antora-playbook.yaml` actually makes sense. To include the tags I'd change the checkout action to check out _all_ of the history, instead of just the latest commit (which is the default). This is also a simple change in the action file.

Acceptance criteria:
- [ ] deploy action file adapted to use other playbook
- [ ] build action file adapted to use full commit history
- [ ] online docs show 22.06 and nightly version

siegfriedweber:
Fixes #269 

The solution described in #269 is implemented.
This pull request will be postponed until the release 22.09 is done.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ jobs:
       - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
         with:
           submodules: true
+          fetch-depth: 0
       - uses: actions/setup-node@1f8c6b94b26d0feae1e387ca63ccbdc44d27b561 # tag=v2
         with:
           node-version: '16'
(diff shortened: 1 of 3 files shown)


fhennig (approved):
Thanks for the playbook fixes

LGTM

PR #275, opened 2022-09-12
Added tracking keys

fhennig:
Tracking site keys which are read by the documentation UI. Tracking is only enabled in the production playbook.


stackfab (approved):
Looks good.

PR #292, opened 2022-09-28
Build documentation for listener-operator

teozkr:


maltesander (approved):
LGTM.

lfrancke:
Forgive my ignorance but: https://docs.stackable.tech/listener-operator/stable/listener.html
This currently doesn't have the version selection dropdown as others do.
Will this happen at the next release?
@teozkr @fhennig

teozkr:
I noticed the same thing, and as I understand it that should be the case...

lfrancke:
I talked to Felix and he agrees, thank you!

PR #301, opened 2022-10-26
Logging ADR

fhennig:
Fixes https://github.com/stackabletech/issues/issues/261


modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,153 @@
+= Logging and Log Aggregation Architecture
+Felix Hennig <felix.hennig@stackable.tech>
+v0.1, 2022-10-26
+:status: [draft]
+
+* Status: {status}
+* Deciders: Stackable Team
```
chaoran-chen, line 7: "Stackable Team" is not very specific for a long-lasting document. It might be clear to everyone on the team at the moment (because everyone was involved) but the team can change. In a couple of years, a newcomer might not know who was involved in the decision process.

fhennig: fixed, thanks ;)

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+https://opensearch.org/[OpenSearch] has been picked as the sink for all the aggregated logs. No other opens have been considered, but as the sink sits at the very top of the chain, it is easy to swap it out for a different software.
+
+We want to make it easy for users to deploy OpenSearch, but also easy to switch out the sink.
+
+(How and where is this configured, and how can we make it easy for the user?)
+
+=== Logging configuration
+
+**Log levels** - It is common to have four or five log levels: Error, Warning, Info, Debug and - optionally - Trace. We want to support them across all products. The **default log level** should be Info.
```
siegfriedweber, line 78: Filtering of aggregated logs is much more convenient if the range of log levels is the same for all products. So I would exactly define the emitted "Stackable" log levels here:

```suggestion
**Log levels** - The following log levels are supported: ERROR, WARN, INFO, DEBUG, and TRACE. We support them across all products. The **default log level** is INFO.
```

The log levels should be mapped accordingly by the log agent. For instance, Superset emits logs with the levels CRITICAL, ERROR, WARNING, INFO, and DEBUG. The mapping should be:

| Superset log level | Stackable log level |
| --- | --- |
| CRITICAL | ERROR |
| ERROR | ERROR |
| WARNING | WARN |
| INFO | INFO |
| DEBUG | DEBUG |
| DEBUG | TRACE |

There is no TRACE log level in Superset, so if the user sets the desired log level to TRACE then it is actually set to DEBUG in Superset.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,161 @@
...
+    roleGroups:
+      default:
+        logging:
+          ...
+      aDifferentGroup:
+        logging:
+          ...
+```
+
+**Override everything** - The customer should be able to supply their own configuration file. Where this is placed depends on the product.
```
siegfriedweber, line 119: It should be mentioned that it is possible to provide a custom configuration file **per role and role-group**.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+```
+logging:
+  custom:
+    configMap: nameOfMyConfigMapWithTheConfigFile
+```
+
+Setting the `custom` field will disable any configurations made in `file` and `stdout`.
```
siegfriedweber, line 127: Should we disable or disallow contradicting configurations? "Disable" means: Accept the configuration but emit a warning. "Disallow" means: Reject the configuration.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+**Disable vector** - Vector should be optional, if the user wants to use their own logging system.
+
+```
+logging:
+  enableVectorAgent: false  # defaults to true
```
siegfriedweber, line 133: This configuration option makes only sense at the top level and not per role or role-group. So the global and non-global options differ.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,161 @@
...
+```
+
+=== Deploying the Stack
+
+The operator deploys the Vector agent as a sidecar and deploys the logging configuration for the product.
+
+The aggregator and OpenSearch sink are deployed with Helm for now, with a plan to integrate this into stackablectl. _Maybe_ we build our own operators for Vector and OpenSearch in the future.
```
siegfriedweber, line 140: I already created a logging stack in stackablectl which deploys OpenSearch, OpenSearch Dashboards, and a Vector log aggregator.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
fhennig: Hmm, are you sure? Every role-group is its own StatefulSet and so we can decide on a per role/group level if the sidecar should be added or not, can't we?

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
fhennig: nice. Added that.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
siegfriedweber: You are right.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,161 @@
...
+As a user of the platform, I have poor visibility into what is happening in different components of the platform, as well as how components influence each other, especially the operator and the product clusters it manages.
+
+Logs usually give this insight. Currently, every pod logs to stdout with a certain default log format. For log configuration we support setting a custom log configuration file in some products.
+
+Logs are **not persisted**, a crashed pod is difficult to investigate. Logs are **not aggregated**, investigating issues that involve multiple pods is difficult. Log **configuration is difficult** or impossible.
+
+== Decision Drivers
+
+* **identical configuration** - All product logs should be configurable in the same way, no matter their underlying logging framework (log4j, logback, Python logging, etc.). Every product should have a logging section in its CRD.
+* **easily configured sinks** - A log sink (i.e. elasticsearch) should only be configured in a single place for the whole platform (not in every product instance).
```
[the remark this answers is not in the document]
fhennig: We're doing a discovery configmap

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+https://opensearch.org/[OpenSearch] has been picked as the sink for all the aggregated logs. No other options have been considered, but as the sink sits at the very top of the chain, it is easy to swap it out for a different software.
+
+The sink is configured in the Vector aggregator. We initially do not provide any "Stackable way" of configuring the sink, but it is easily configured by the user in a single location for the whole platform. Other sinks can be configured easily as well (https://vector.dev/docs/reference/configuration/sinks/[Vector docs for sink configuration]). See also <<deploying_the_stack>>.
+
+**For now, setting up the aggregator and sink is the responsibility of the user**.
+
+=== Logging configuration
+
+**Log levels** - The following log levels are supported: ERROR, WARN, INFO, DEBUG, and TRACE. We support them across all products. The **default log level** is INFO. Products might have slightly different log levels, which are mapped accordingly by the agent. For instance, Superset emits logs with the levels CRITICAL, ERROR, WARNING, INFO, and DEBUG. The mapping should be:
```
siegfriedweber, line 81: Logback and log4j have the following log levels: TRACE, DEBUG, INFO, WARN, ERROR, **FATAL**, **OFF**
We should support the same. Then we can map the CRITICAL level which is used by Python to FATAL. Turning off logs for some loggers could also be useful.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,196 @@
...
+
+=== Log Sink: OpenSearch
+
+https://opensearch.org/[OpenSearch] has been picked as the sink for all the aggregated logs. No other options have been considered, but as the sink sits at the very top of the chain, it is easy to swap it out for a different software.
+
+The sink is configured in the Vector aggregator. We initially do not provide any "Stackable way" of configuring the sink, but it is easily configured by the user in a single location for the whole platform. Other sinks can be configured easily as well (https://vector.dev/docs/reference/configuration/sinks/[Vector docs for sink configuration]). See also <<deploying_the_stack>>.
+
+**For now, setting up the aggregator and sink is the responsibility of the user**.
+
+=== Logging configuration
```
siegfriedweber, line 79: The log configuration is currently described per pod but it must actually be done per container. So the configuration should be structured as follows:

```
spec:
  vectorAggregatorConfigMapName: ...
  role:
    roleGroups:
      myFirstRoleGroup:
        config:
          logging:
            enableVectorAgent: true
            containers:
              myFirstContainer:
                console:
                  levelThreshold: INFO
                file:
                  levelThreshold: WARN
                loggers:
                  ROOT:
                    level: INFO
                  another.logger:
                    level: ERROR
              mySecondContainer: ...
```

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
```diff
@@ -0,0 +1,196 @@
...
+        level: DEBUG
+  console:
+    loggers:
+      ROOT:
+        level: DEBUG
+      my.specific.package:
+        level: TRACE
+      my.other.package:
+        level: TRACE
+```
```
siegfriedweber, line 128: `file` and `console` are called appenders in logback. They are on the same level as loggers. Therefore the given structure cannot be directly translated to logback or log4j. I would suggest the following structure:

```
logging:
  containers:
    myContainer:
      console:
        levelThreshold: INFO
      file:
        levelThreshold: WARN
      loggers:
        ROOT:
          level: INFO
        another.logger:
          level: ERROR
```

The log level per logger can be defined in `loggers`. The appenders only allow to set a threshold for the log level.

modules/contributor/pages/adr/ADRXYZ_logging_architecture.adoc
siegfriedweber: It turned out that `OFF` is recognized as the boolean value `false` by YAML parsers. So `NONE` should be used instead of `OFF`.

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+** Siegfried Weber
+** Sönke Liebau
+** Teo Klestrup Röijezon
+* Date: 2022-12-07
+
+Technical Story: https://github.com/stackabletech/issues/issues/202, https://github.com/stackabletech/issues/issues/261
+
+== Context and Problem Statement
+
+// Describe the context and problem statement, e.g., in free form using two to three sentences. You may want to articulate the problem in form of a question.
```
adwk67, line 22:
```suggestion
// Describe the context and problem statement, e.g., in free-form using two to three sentences. You may want to articulate the problem in form of a question.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+As a user of the platform, I have poor visibility into what is happening in different components of the platform, as well as how components influence each other, especially the operator and the product clusters it manages.
```
adwk67, line 24:
```suggestion
As a user of the platform, I have poor visibility into what is happening in different components of the platform and how these components influence each other, especially operators and the product clusters they manage.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+Logs usually give this insight. Currently, every pod logs to stdout with a certain default log format. For log configuration we support setting a custom log configuration file in some products.
```
adwk67, line 26:
```suggestion
Logs are used to make this information available. Currently, every pod logs to stdout with a certain default log format. For log configuration we support setting a custom log configuration file in some products.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+Logs are **not persisted**, a crashed pod is difficult to investigate. Logs are **not aggregated**, investigating issues that involve multiple pods is difficult. Log **configuration is difficult** or impossible.
```
adwk67, line 28:
```suggestion
However, logs are **not persisted**, meaning that a crashed pod is difficult to investigate. They are also **not aggregated**, so that investigating issues involving multiple pods is difficult. Log **configuration is difficult** and sometimes not even possible.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+== Decision Drivers
+
+* **Identical configuration** - All product logs should be configurable in the same way, no matter their underlying logging framework (log4j, logback, Python logging, etc.). Every product should have a logging section in its CRD.
+* **Easily configured sinks** - A log sink (i.e. elasticsearch) should only be configured in a single place for the whole platform (not in every product instance).
+* **Support plaintext logs on stdout** - It is a Kubernetes convention to log in plaintext to stdout. This should be kept, as it is a useful tool for interactive debugging.
```
adwk67, line 34:
```suggestion
* **Support plaintext logs on stdout** - It is a Kubernetes convention to log in plaintext to stdout. This should be kept as it is a useful tool for interactive debugging.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+* **Custom overrides** - The log format, aggregator and sink should be substitutable by the user.
```
adwk67, line 35:
```suggestion
* **Custom overrides** - The user should be able to override the log format, aggregator and sink.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+* **Transport security** - Logs should transmitted with encryption/TLS.
+
+=== Assumptions
+
+* **RegEx parsing is error prone** - Parsing pure string log output to get structure (time, module, log level, message, ...) is prone to errors. Especially for multi-line messages like stacktraces. The design should use structured log ouput whenever possible.
```
adwk67, line 40:
```suggestion
* **RegEx parsing is error prone** - Parsing pure string log output to get structure (time, module, log level, message, ...) is prone to errors, especially for multi-line messages like stacktraces. The design should use structured log output whenever possible.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+=== Constraints
+
+* **Multiple files** - Some products like HDFS _need_ to log to (multiple) files, so parsing just stdout is not sufficient; we need to parse files.
```
adwk67, line 44:
```suggestion
* **Multiple files** - Some products like HDFS _need_ to log to (multiple) files, so parsing stdout alone is not sufficient; we need to parse files.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+== Decision
+
+We decided to use https://vector.dev/[Vector] as an aggregator. It looked good, no other options were evaluated.
```
adwk67, line 48:
```suggestion
We decided to use https://vector.dev/[Vector] as an aggregator. It looked promising - no other options were evaluated.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+We chose to provide **plaintext logs on stdout** and **structured logs on file**. For the architecture, we chose to deploy the agent as a **sidecar** and **use an aggregator**. We chose to provide **OpenSearch** as a default Sink for logs.
```
adwk67, line 50:
```suggestion
We chose to provide **plaintext logs on stdout** and **structured logs on file**. For the architecture, we chose to deploy the agent as a **sidecar** and to **use an aggregator**. We chose to provide **OpenSearch** as a default Sink for logs.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+image::adr/logging_architecture.drawio.svg[]
+
+Find detailed considerations below.
```
adwk67, line 54:
```suggestion
Detailed considerations are given below.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+=== Log Aggregation Framework: Vector
+
+https://vector.dev/[Vector] has been picked without comparison to other existing choices. Some googling shows https://medium.com/ibm-cloud/log-collectors-performance-benchmarking-8c5218a08fea[Vector has better performance than fluentd], the biggest name in this space. Vector is also open source and written in Rust: https://github.com/vectordotdev/vector[GitHub].
+
+=== Log Retrieval Strategy
+
+We want to parse logs out of structured log entries and also keep the stdout log as plaintext. Also, some products require reading log files instead of stdout. Which means **we opt to read logs from files for every product** and **parse structured log entries**.
```
adwk67, line 62:
```suggestion
We want to parse logs out of structured log entries and also keep the stdout log as plaintext. Some products also require reading log files instead of stdout. Which means **we opt to read logs from files for every product** and **parse structured log entries**.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+[#log_aggregation_architecture]
+=== Log Aggregation Architecture
+
+**Agent role** - Vector offers multiple https://vector.dev/docs/setup/deployment/roles/#agent[roles] in which it can run. The DaemonSet is easier to deploy, and only one vector instance is running per host. But the sidecar has deeper integration with the product container and can read log files, something that we decided we need.
```
adwk67, line 67:
```suggestion
**Agent role** - Vector offers multiple https://vector.dev/docs/setup/deployment/roles/#agent[roles] in which it can run. The DaemonSet is easier to deploy, and only one vector instance is running per host. But the sidecar has deeper integration with the product container and can read log files - something that we decided we need.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+**Topology** - Vector offers multiple https://vector.dev/docs/setup/deployment/topologies/[topologies] and already lists pros and cons. An aggregator makes it easier to configure a sink: only one place to configure it (the aggregator has "multi-host context" as Vector calls it). It also reduces requests made to the sink, as it aggregates and buffers requests.
```
adwk67, line 69:
```suggestion
**Topology** - Vector offers multiple https://vector.dev/docs/setup/deployment/topologies/[topologies] and already lists pros and cons. An aggregator makes it easier to configure a sink as there is only one place to configure it (the aggregator has "multi-host context" as Vector calls it). It also reduces requests made to the sink, as it aggregates and buffers requests.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+The connection from the agent to the aggregator is made with the help of a **discovery ConfigMap**. The aggregator is discoverable with a discovery ConfigMap, which is referenced in the ProductCluster with a property called `vectorAggregatorConfigMapName`, similar to `zookeeperConfigMapName` or `hdfsConfigMapName`. The property is located at the top level of the spec.
```
adwk67, line 71:
```suggestion
The connection from the agent to the aggregator is made with the help of a **discovery ConfigMap**. The aggregator can be detected with a discovery ConfigMap, which is referenced in the ProductCluster with a property called `vectorAggregatorConfigMapName`, similar to `zookeeperConfigMapName` or `hdfsConfigMapName`. The property is located at the top level of the spec.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+=== Log Sink: OpenSearch
+
+https://opensearch.org/[OpenSearch] has been picked as the sink for all the aggregated logs. No other options have been considered, but as the sink sits at the very top of the chain, it is easy to swap it out for a different software.
```
adwk67, line 75:
```suggestion
https://opensearch.org/[OpenSearch] has been picked as the sink for all the aggregated logs. No other options have been considered, but as the sink sits at the very top of the chain, it is easy to swap it out for different software.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+=== Logging Configuration
+
+**Log levels** - The following log levels are supported: NONE, FATAL, ERROR, WARN, INFO, DEBUG, and TRACE. We support them across all products. The **default log level** is INFO. Products might have slightly different log levels, which are mapped accordingly by the agent. For instance, Superset emits logs with the levels CRITICAL, ERROR, WARNING, INFO, and DEBUG. The mapping should be:
+
+[cols="1,1"]
+|===
+|Superset log level | Stackable log level 
+
+|CRITICAL
```
adwk67, line 89:
```suggestion
| CRITICAL
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+    roleGroups:
+      default:
+        logging:
+          ...
+      aDifferentGroup:
+        logging:
+          ...
+```
+
+**Configuration per container** - While we don't typically configure things at the container level, it is necessary to do so for logging. As shown below we want to be able to set log levels for specific modules or override a log configuration file entirely. This is however container specific. For example, an init container, the product container itself and the vector container are all configured in different ways, and offer different modules for which log level can be set. And because of that, log configuration needs to be specified per container.
```
adwk67, line 129:
```suggestion
**Configuration per container** - While we don't typically configure things at the container level, it is necessary to do so for logging. As shown below we want to be able to set log levels for specific modules or override a log configuration file entirely. This is however container specific. For example, an init container, the product container itself and the vector container are all configured in different ways, and offer different modules for which the log level can be set. For this reason log configuration needs to be specified per container.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+```yaml
+logging:
+  containers:
+  myContainer:
+    custom:
+      configMap: nameOfMyConfigMapWithTheConfigFile
+```
+
+Like the other logging settings, this costum configuration file can be supplied per role and/or role-group.
```
adwk67, line 190:
```suggestion
Like the other logging settings, this custom configuration file can be supplied per role and/or role-group.
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+Setting the `custom` field will disable any configurations made in `file` and `console`. (TODO maybe we can disallow this alltogether in the CRD type)
```
adwk67, line 192:
```suggestion
Setting the `custom` field will disable any configurations made in `file` and `console`. (TODO maybe we can disallow this altogether in the CRD type)
```

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,231 @@
...
+
+```yaml
+logging:
+  enableVectorAgent: true
+  containers:
+    myFirstContainer:
+      console:
+        levelThreshold: INFO
+      file:
+        levelThreshold: WARN
```
adwk67, line 177: I find this example a bit confusing and/or incomplete: why do we have "level" for modules but "levelThreshold" for containers? What else can be configured here other than level? (or will that be documented elsewhere?)

fhennig: good point, I'll go ask Sigi

siegfriedweber: The naming is inspired by log4j and logback.

```
loggers:
  my.module:
    level: INFO
```

is translated to

```
<logger name="my.module" level="INFO" />
```

And

```
console:
  levelThreshold: INFO
```

is translated to

```
<appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
  <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
    <level>INFO</level>
  </filter>
</appender>
```

For now, nothing else can be configured at this level. But having a "level" field allows extending the structure and is self-descriptive:

```
console: INFO
```

vs.

```
console:
  levelThreshold: INFO
```

adwk67: Does this mean the user is implicitly configuring log4j when using `level: INFO` and logback when using `levelThreshold: INFO`?

siegfriedweber: No, the configuration is independent of the logging framework. The structure is as follows:

```
logging:
  enableVectorAgent: true
  containers:
    myContainer:
      console:
        levelThreshold: INFO
      file:
        levelThreshold: INFO
      loggers:
        ROOT:
          level: INFO
        my.module:
          level: INFO
      custom:
        configMap: my-configmap
```

If the product uses log4j then this configuration is translated into a log4j configuration. If it uses logback then the configuration is translated into a logback configuration. The same goes for any other logging framework.

It would be okay for me to rename `levelThreshold` into `level` if that avoids confusion.

adwk67: Thanks for the explanation! I would personally go for `level` everywhere as I find that less confusing.

fhennig: I've made a change to use `level` instead of `levelThreshold` in the ADR

modules/contributor/pages/adr/ADR025-logging_architecture.adoc
```diff
@@ -0,0 +1,253 @@
...
+  containers:
+    myFirstContainer:
+      loggers:
+        ROOT:
+          level: INFO
+        another.logger:
+          level: ERROR
+```
+
+The ROOT logger is not tied to a module, but is configures the overall log level of the underlying logging framework.
```
adwk67, line 167:
```suggestion
The ROOT logger is not tied to a module but configures the overall log level of the underlying logging framework.
```

adwk67 (approved):
Very nice! LGTM (just one minor change)

adwk67 (approved):

Issue #299, opened 2022-10-10
Fix link to Trino authorization


siegfriedweber:
In stackabletech/trino-operator#291, the Trino documentation was restructured and now the link to Trino in https://docs.stackable.tech/home/stable/concepts/opa.html#_further_reading does not work anymore and must be fixed.

The build process shows the following errors:

```
[10:55:03.775] ERROR (asciidoctor): target of xref not found: trino::usage.adoc#_authorization
    file: modules/concepts/pages/opa.adoc
    source: https://github.com/stackabletech/documentation.git (refname: docs/22.06)
[10:55:07.725] ERROR (asciidoctor): target of xref not found: trino::usage.adoc#_authorization
    file: modules/concepts/pages/opa.adoc
    source: https://github.com/stackabletech/documentation.git (refname: docs/22.09)
[10:55:09.636] ERROR (asciidoctor): target of xref not found: trino::usage.adoc#_authorization
    file: /home/sigi/projects/stackable/documentation/modules/concepts/pages/opa.adoc
    source: /home/sigi/projects/stackable/documentation (refname: main <worktree>)
```

The links point to the latest stable versions of the operators, so the documentation of all releases is affected which should also not be the case.

fhennig:
Yes. This problem might also go beyond this link. To prevent this from happening, all links should be versioned as well, and link to a specific version of a component (operator). This might mean, that when we make new releases we also need to update versions in links. Very painful.

https://docs.antora.org/antora/latest/navigation/xrefs-and-link-text/

teozkr:
I split off the more general link versioning into #302 since I think that's a bigger issue that we should definitely solve, but which I don't want to let block the more immediate issue that _our docs are currently broken_.

lfrancke:
Thanks Teo!

fhennig:
Good idea to move the general problem to a different issue.

For the issue at hand: As Sigi said, all releases are affected. The fix is to update the link to point to either an older version of the Trino operator, or the new authorization page (https://docs.stackable.tech/trino/stable/usage_guide/security.html#_authorization).

Udating nightly should be straightforward. Updating the 22.06 and 22.09 is more difficult because those are pulled from tags. I think we'd have to make branches, fix the link (pointing to a different version of the trino operator) and move the tag to a new commit. It would then not be in main anymore.

teozkr:
[...]

PR #303, opened 2022-11-01
Fix link to trino authorization

teozkr:
See #299


fhennig (dismissed):
This fixes nightly

fhennig (approved):
looks good, but should only be merged once the branches exist

the branches should link to the correct trino version

PR #305, opened 2022-11-01
Lock Trino authz link to 0.6

teozkr:
See #299


fhennig (approved):

fhennig:
I think the build failure is caused by multiple branches/tags having the same version in the antora.yml file. At least I have seen this caused by that. In this case, it is probably HEAD and the docs/22.09 tag both having the "22.09" version, which confuses antora (since antora only looks at its internal version of a component)

teozkr:
Yeah.. I was running into that locally too. :/

PR #307, opened 2022-11-03
Search fix

### LINKED ISSUE #279: Search returns broken links
~~A search for “Hive” on https://docs.stackable.tech/home/stable/index.html the first entry isn’t clickable.~~ Clicking the link going to the main page doesn't jump to the correct section on the page.

Screenshot 2022-09-13 at 11 36 24

## Expected behaviour

Either a jump to the [`_components` anchor](https://docs.stackable.tech/home/stable/index.html#_components) or to a specific "Hive" anchor on the page

## Acceptance criteria

- [ ] This is fixed when the search causes the jump to the section as described above
- [ ] Change should be applied to all releases

fhennig:
fixes #279 in main

I've tried the regular anchor syntax for antora, but it didn't work. I suspect, because we made the heading with html instead of adoc. So I made the anchor also in html.

I've tested it, and search for hive will no jump to the right place.

### DIFF
--- a/modules/ROOT/pages/index.adoc
+++ b/modules/ROOT/pages/index.adoc
@@ -87,7 +87,7 @@ The Stackable Data Platform supports the following products:
 ++++
 
 ++++
-<h3>Apache Airflow</h3>
+<h3 id="airflow"><a class="anchor" href="#airflow"></a>Apache Airflow</h3>
 ++++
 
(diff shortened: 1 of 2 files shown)


fhennig:
Ooops, the logging adr commits got in here too. closing in favor of #308

PR #308, opened 2022-11-03
Added refs to the h3 headings to fix search

fhennig:
fixes https://github.com/stackabletech/documentation/issues/279 in main

I've tried the regular anchor syntax for antora, but it didn't work. I suspect, because we made the heading with html instead of adoc. So I made the anchor also in html.

I've tested it, and search for hive will no jump to the right place.


razvan (approved):
lgtm. tested locally.

PR #325, opened 2022-12-12
concepts reorganized and new overview page

fhennig:
I made this during the break week


fhennig:
Look here: https://deploy-preview-325--stackable-docs.netlify.app/home/nightly/concepts/

modules/concepts/images/operator_overview.drawio.svg
sbernauer, line 1: Nit: STS and CM are actually per role group, not role

modules/concepts/pages/overview.adoc
```diff
@@ -0,0 +1,68 @@
...
+Learn more about roles: xref::roles-and-role-groups.adoc[]
+
+[#deployment]
+== Deployment
+
+All operators and products run as containers in a Kubernetes cluster. The operators are deployed with stackablectl (the Stackable CLI) or Helm.
+
+image::deployment.drawio.svg[]
+
+To deploy a product instance, a product resource needs to be created in Kubernetes, this is usually done by passing a YAML manifest file to kubernetes with `kubectl apply -f`. The manifest file contains the configuration of how the product should operate.
```
sbernauer, line 53:
```suggestion
To deploy a product instance, a product resource needs to be created in Kubernetes, this is usually done by passing a YAML manifest file to kubernetes with `kubectl apply -f <file.yaml>`. The manifest file contains the configuration of how the product should operate.
```

modules/concepts/images/common_objects.drawio.svg
sbernauer, line 1: Can we please rename `LdapServer` to `Ldap configuration`? Technically AuthenticationClass would be correct, but that could be considered a implementation detail

sbernauer (changes requested):
Awesome. very nice!

modules/concepts/images/operator_overview.drawio.svg
fhennig: As there are no role groups in the image at all, I'd just leave it like this. I think the general concept is clear, and that's what this illustration is for

modules/concepts/images/common_objects.drawio.svg
fhennig: I fixed this one! Since S3Bucket is an actual object I turned the Ldap concfig into an AuthenticationClass

modules/concepts/images/operator_overview.drawio.svg
sbernauer: Alright

sbernauer (approved):

PR #337, opened 2023-01-10
More version bumps

fhennig:


maltesander (approved):
LGTM if build is green

PR #338, opened 2023-01-10
Correction to example NifiCluster - moving 'selector' out of 'config'

pipern:
According to https://github.com/stackabletech/nifi-operator/blob/main/examples/simple-nifi-cluster.yaml `selector` should be within `nodes.roleGroups.default` not `nodes.roleGroups.default.config`


fhennig (dismissed):
Looks good, thanks for spotting that!

fhennig:
Whoops! I didn't want to mess with your branches. The selector shouldn't be needed, I'll remove it upstream but I'll use a different PR because it needs to be merged into main.

fhennig:
fixed here: https://github.com/stackabletech/documentation/commit/303f30f8c76e1343c1d49e868643899cba3da58c

PR #362, opened 2023-02-14
Add ADR 26 on affinities

sbernauer:
For https://github.com/stackabletech/issues/issues/323


soenkeliebau (approved):
Approved based on @sbernauer reading and approving my commit to the branch

PR #366, opened 2023-02-28
ADR 27 status field

razvan:
todo

### DIFF
--- a/modules/contributor/pages/adr/ADR027-status.adoc
+++ b/modules/contributor/pages/adr/ADR027-status.adoc
@@ -1,0 +1,260 @@
+= [Affinities]
+Sebastian Bernauer <sebastian.bernauer@stackable.tech>
+v0.1, 2023-02-13
+:status: draft
+
+* Status: {status}
+* Deciders:
+** Sönke Liebau
+** Razvan Mihai
+** Sebastian Bernauer
+* Date: 2023-02-13
+
+Technical Story: https://github.com/stackabletech/issues/issues/323
+
+== Context and Problem Statement
+
+// Describe the context and problem statement, e.g., in free form using two to three sentences. You may want to articulate the problem in form of a question.
+
+When running multiple instances of services of a data product most of the time it makes sense to influence the way Pods get assigned to Nodes.
+In some cases it makes sense to co-locate certain services that talk a lot to each other, such as HBase regionservers with HDFS datanodes.
+In other cases it makes sense to distribute the Pods among as much Nodes as possible.
+There can also be some additional requirements, such as placing important services - such as HDFs namenode - in different racks, datacenter rooms or even datacenters.
+
+This ADR proposes a solution to automatically deploy some default affinities that should work for most users out-of-the box and improve the availability of the products.
+Additionally users need to be able to configure their own affinity rules on a role as well as role-group level.
+
+== Decision Drivers
+
+During our deliberations we worked out the following main use cases that should be possible with the chosen solution:
+
+1. Leave defaults as set by the operator -> no nodeAffinity oder nodeSelect, podAffinities
+2. Set node affinities, but leave the default pod affinities as set by the operator
+3. Override pod affinities set by the operator with custom ones
+4. Override pod affinities and set node affinities at the same time
+
+In addition to these use-cases, our operators currently offer a `nodeSelector` field in the CRD, which offers similar functionality to the node affinities discussed in this ADR.
+The chosen option needs to enable us to properly handle the existing field going forward and have a defined migration path towards deprecating this field and using the more detailed node affinities.
+
+== Considered Options
+
+=== Use podOverwrite
+
+Don't handle affinities in a dedicated attribute, but let the users use `podOverwrite` when it is implemented https://github.com/stackabletech/issues/issues/346[tracked by this Issue]
+
+==== Pros
+
+* No extra implementation effort once `podOverwrite` has been done
+* No CRD changes needed
+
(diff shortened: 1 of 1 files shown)


modules/contributor/pages/adr/ADR027-status.adoc
```diff
@@ -0,0 +1,176 @@
...
+* True
+* False
+* Unknown
+
+== Decision Drivers
+
+All Stackable Operators should update the health status of cluster resources and sub-resources they manage (where this makes sense). For example, The Stackable Operator for Apache Zookeeper should manage the health status of `ZookeeperCluster` as well as `ZNode` resources.
+
+Must-haves for cluster resources and sub-resources:
+- Status string field to allow easy programmatic query of the resource state. This should be one of the condition types recommended above (Available, Stopped, Paused, ...).
```
maltesander, line 46: I think this is not a must have anymore?

maltesander (approved):
Just one question. LGTM otherwise!

modules/contributor/pages/adr/ADR027-status.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+For example, the Superset operator should query the StatefulSets, database jobs, etc. and aggregate them into cluster conditions of the `SupersetCluster`.
+
+In the simplest case, if all sub-resources have a health type of "Ready" (as usually defined by Kubernetes) with status the `True`, then these will be aggregated into a condition of type "Available" with the status "True" at the product cluster level.
+
+In often times however, sub-resources might be transitioning from one phase to another. In such cases, the product cluster state might still have a condition "Available" set to true but also an additional "Degraded" condition with the status "True". This might reflect that even if the product and it's services are available and running, the entire cluster setup does not correspond to the requested custom resource.
+
+
+QUESTION: should dependent services be considered for status checks ? For example, should the Superset operator check that Druid services function properly ?
```
fhennig, line 72: Good question. I think if i.e. ZooKeeper doesn't work anymore and Kafka needs zookeeper, obviously Kafka is affected.

But if Druid is degraded but still works, it's difficult to say whether Superset is degraded or not. I guess if we don't want to look into the specifics of _what_ is degraded in Druid, we just can't take it into account anways.

modules/contributor/pages/adr/ADR027-status.adoc
```diff
@@ -0,0 +1,176 @@
...
+==== Pros
+
+* Implementation detail: reuse `ClusterResources` from the `operator-rs` framework as much as possible and have a generic way to update cluster status.
+* Non-breaking CRD changes needed to add conditions.
+* Transparent and easy to understand. Failure messages can be propagated from sub-resources hat have problems.
+
+
+==== Cons
+
+* It completely relies on Kubernetes resource status fields without querying the actual products. This assumes that liveliness probes and other checks that Kubernetes performs, mirror the true product status.
```
fhennig, line 83: I don't think this is a Con, this is good! We should use k8s features whereever we can instead of reinventing them. If the liveliness probes suck we should fix them.

modules/contributor/pages/adr/ADR027-status.adoc
```diff
@@ -0,0 +1,176 @@
...
+* It requires that all resource dependencies run inside the Kubernetes cluster. For example, if a Superset cluster is configured with a Druid connection outside the Kubernetes cluster, the `Available` condition of the connection will have a status of `Unknown`. The Superset cluster itself, might have a "Degraded" condition with status "True" and a message informing the user that the Druid connection cannot be queried.
```
fhennig, line 84: This is a problem regardless of how the Status fields are structured, so it's fine

modules/contributor/pages/adr/ADR027-status.adoc
```diff
@@ -0,0 +1,176 @@
...
+    - type: Stopped
+      status: "True"
+      lastProbeTime: 2023-02-28T14:02:00Z
+      lastTransitionTime: 2023-02-28T12:00:00Z
+      reason: "User requested reconcile stop."
+----
+
+=== Set Status Custom Fields and Conditions
+
+Most custom fields are set by querying the products directly. One exception is the deployed product version.
```
fhennig, line 138: I don't fully understand what is "on top" of the other solution. Is it just the version? The version is already in the spec isn't it?

fhennig (changes requested):
A few formatting things, but looks good overall!

Also I had a few questions for clarification.

modules/contributor/pages/adr/ADR027-status.adoc
maltesander: In general you are right we should fix the liveness/readiness probes as well. But there are some health checks that can only be performed with a certain protocol talking to the product directly that cannot be done with Socket-/TCP/HTTP probes. 
So at least some of our operators need to be able to communicate with their products to determine their "health". E.g. HDFS operator and under-replicated block count.

modules/contributor/pages/adr/ADR027-status.adoc
razvan: Superset would be both degraded (true) and available (true) without the need to know  why Druid is degraded.

But in the first implementation of this ADR, we decided to ignore the status of such dependencies.

modules/contributor/pages/adr/ADR027-status.adoc
fhennig: Ahh I see

fhennig (approved):

fhennig (approved):

PR #392, opened 2023-04-24
Build from CI cloned repo (and fix netlify previews)

fhennig:
This PR changes the production build to build from the locally fetched repo in the CI.

As the build happens in netlify, a script is necessary to fetch all to content - netlify does a bare clone initially.

As a consequence the netlify previews in PRs now actually work!


netlify.toml
```diff
@@ -1,3 +1,7 @@
+[build]
+  command = "bash scripts/netlify_git_fetch.sh && make ANTORAFLAGS=--fetch PLAYBOOK=antora-playbook.yml"
```
razvan, line 2: Why not  put the content of the script in the Makefile ?

razvan (commented):
one question

netlify.toml
fhennig: Good question. I'll try to do that, would be cleaner

fhennig: I did it!

razvan (approved):
nice

fhennig:
lovely jubbly!

PR #402, opened 2023-05-11
Don't suggest that users delete CRDs when upgrading

nightkr:
See https://stackable-workspace.slack.com/archives/C031BTE7ZGC/p1683797224359929

Cherry-pick of #405

### DIFF
--- a/modules/ROOT/pages/release_notes.adoc
+++ b/modules/ROOT/pages/release_notes.adoc
@@ -271,39 +271,34 @@ $ stackablectl release uninstall 23.1
 # ...
 ----
 
-Afterwards you will need to delete the CustomResourceDefinitions (CRDs) installed by the Stackable Platform.
-The reason for this is that helm will uninstall the operators but not the CRDs. This can be done in two ways:
-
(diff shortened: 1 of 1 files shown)


razvan (approved):

sbernauer (approved):
Thanks!

nightkr:
@fhennig did anything change lately regarding the Antora build?

nightkr:
So this is causing existing objects to become impossible to modify after the upgrade is complete. This is because the new field `status.conditions` is mandatory, but doesn't have a default value. This would have to be fixed in each operator, by adding a default. For example, in nifi-operator:

```diff
diff --git a/rust/crd/src/lib.rs b/rust/crd/src/lib.rs
index 402f8bc..0b8e6b7 100644
--- a/rust/crd/src/lib.rs
+++ b/rust/crd/src/lib.rs
@@ -205,6 +205,7 @@ pub enum NifiRole {
 #[derive(Clone, Debug, Default, Deserialize, JsonSchema, Serialize)]
 pub struct NifiStatus {
     pub deployed_version: Option<String>,
+    #[serde(default)]
     pub conditions: Vec<ClusterCondition>,
 }
```

nightkr:
Ok, this has now been updated from #405.

adwk67 (approved):

PR #404, opened 2023-05-12
Update breaking change notes on Airflow

razvan:


modules/ROOT/pages/release_notes.adoc
```diff
@@ -101,6 +101,12 @@ spec:
 ```
 This is an example for Trino, but the pattern is the same across all operators.
 
+==== Stackable Operator for Apache Airflow
+
+Existing Airflow clusters need to be deleted and recreated.
```
adwk67, line 106: Maybe add a note on the fact that data is not lost?
```suggestion
Existing Airflow clusters need to be deleted and recreated. Airflow metadata held in the database and DAGs saved on disk are not affected.
```

adwk67 (approved):
lgtm + one minor suggestion.

PR #415, opened 2023-06-13
Document user overrides

sbernauer:
For https://github.com/stackabletech/issues/issues/346


modules/concepts/pages/user_overrides/config_overrides.adoc
```diff
@@ -0,0 +1,32 @@
+= Config overrides
+
+WARNING: Overriding certain configurations can lead to faulty clusters. User overrides should only be used as a last resort!
```
lfrancke, line 3:
```suggestion
WARNING: Overriding certain properties can lead to faulty clusters. User overrides should only be used as a last resort!
```

modules/concepts/pages/user_overrides/env_overrides.adoc
```diff
@@ -0,0 +1,32 @@
+= Env overrides
```
lfrancke, line 1:
```suggestion
= Environment variable overrides
```

modules/concepts/pages/user_overrides/env_overrides.adoc
```diff
@@ -0,0 +1,32 @@
...
+        fs.trash.interval: "5"
+    roleGroups:
+      default:
+        config: # [..]
+        envOverrides: # on rolegroup level, takes precedence over the ones specified at role level
+          MY_ENV_VAR: "MY_VALUE"
+        replicas: 1
+----
+
+The env vars you can set dependents on the individual product.
```
lfrancke, line 31:
```suggestion
The environment variables you can set depend on the specific product.
```

modules/concepts/pages/user_overrides/index.adoc
```diff
@@ -0,0 +1,5 @@
+= User overrides
+
+WARNING: Overriding certain configurations can lead to faulty clusters. User overrides should only be used as a last resort!
```
lfrancke, line 3:
```suggestion
WARNING: Overriding certain properties can lead to faulty clusters. Overrides should only be used as a last resort!
```

lfrancke (changes requested):

modules/concepts/pages/user_overrides/index.adoc
lfrancke: I'm not sure if that is correct english.
@Jimvin can you help out?

Jimvin: LGTM. It conveys the correct message, perhaps "Overrides should be used with caution!" might be better.

fhennig:
Great, thanks for documenting this! 

One thing: I'd like to have all of these in a single page `overrides.adoc` with 3 sub headings. You can still link to just Pod overrides by linking to the heading. I think they are all conceptually very similar and the pages are quite short.

sbernauer:
@fhennig was thinking about that as well in the beginning. I don't know if we will have more overrides in the future (e.g. for Services and pvcs). That was my main motivation of putting them in dedicated files. WDYT?

modules/concepts/pages/user_overrides/index.adoc
lfrancke: Sorry, but...what looks good to you?

"configurations"? I'm not a native speaker so I might be wrong but it sounds...wrong to me.

Jimvin: They are configuration properties, so either would work. I personally think properties makes more sense but I would be happy with either.

fhennig:
We talked over Slack and I convinced Sebastian of going with a single file

modules/concepts/pages/user_overrides/index.adoc
sbernauer: Now we only have one warning at the top of the page left
It currently states `WARNING: Overriding certain configurations can lead to faulty clusters. Overrides should only be used as a last resort!`
I would say `configurations` includes config, env and pod the best, but happy to be overruled :) WDYT?
An alternative could also be `WARNING: Overriding certain aspects of your cluster setup can lead to faulty clusters. Overrides should only be used as a last resort and with care!`

modules/concepts/pages/overrides.adoc
```diff
@@ -0,0 +1,110 @@
+= Overrides
+
+WARNING: Overriding certain configurations can lead to faulty clusters. Overrides should only be used as a last resort!
```
lfrancke, line 3: I still think it should be either `configuration properties` or just `properties`

`configurations` does not sound correct to me in this context.
The plural of `configuration` is used differently afaik.

lfrancke (dismissed):

modules/concepts/pages/overrides.adoc
sbernauer: Alright, changing to `configuration properties`

modules/concepts/pages/overrides.adoc
```diff
@@ -0,0 +1,110 @@
+= Overrides
+
+WARNING: Overriding certain configuration properties can lead to faulty clusters. Overrides should only be used as a last resort!
+
+The cluster definitions also supports overriding configuration aspects, either per role or per role group, where the more specific override (role group) has precedence over the less specific one (role).
```
fhennig, line 5: I have a suggestion for an introduction to the whole topic:

"The Stackable Operators configure the products they are operating with sensible defaults, and required settings to enable connectivity and securty when enabled. Other important settings are usually exposed as settings in the resource definition. In some cases however, you might want to set certain settings that are not exposed, or override settings that the operator has set on the product.

The resource definitions of all products support overrides, specificially for the product configuration, environment variables, and the Pod spec the Operators generate."

I'd add this before the warning as an introduction to the topic

modules/concepts/pages/overrides.adoc
```diff
@@ -0,0 +1,110 @@
...
+        configOverrides: # on rolegroup level, takes precedence over the ones specified at role level
+          hdfs-site.xml:
+            dfs.namenode.num.checkpoints.retained: "3"
+        replicas: 1
+----
+
+The role, as well as the configuration file and configuration settings available depend on the specific product.
+All override property values must be strings.
+The properties will be formatted and escaped correctly into the file format used by the product.
+
```
fhennig, line 37: For i.e. `.properties` files this is obvious, but I'm wondering how this works for XML. Maybe a quick example could be added?

modules/concepts/pages/overrides.adoc
```diff
@@ -0,0 +1,110 @@
...
+    roleGroups:
+      default:
+        config: # [..]
+        envOverrides: # on rolegroup level, takes precedence over the ones specified at role level
+          MY_ENV_VAR: "MY_VALUE"
+        replicas: 1
+----
+
+The environment variables you can set depend on the specific product.
+All override property values must be strings.
```
fhennig, line 67: Is this correct? I think I can set _any_ variable, but maybe not everything makes sense. But there's not technical limitation, right?

modules/concepts/pages/overrides.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+== Pod overrides
+
+For a role or role group, at the same level of `config`, you can specify `podOverrides` for any of the attributes you can configure on a Pod.
```
fhennig, line 71: "... These overriden settings will be added to the PodTemplate of the StatefulSet of the role group" (link to the roles-and-role-groups concept page)

fhennig (commented):
Some ideas

modules/concepts/pages/overrides.adoc
sbernauer: It can also be daemonsets. Good question is how we want to handle jobs we write out, e.g. for Airflow and Superset.
Will link to the concept

modules/concepts/pages/overrides.adoc
sbernauer: Not sure if I did this right, as the example above is using xml :thinking:

sbernauer:
@fhennig created https://github.com/stackabletech/documentation/pull/416

PR #418, opened 2023-06-27
Adds various policies we will try to follow in the future.

lfrancke:
Enterprise Readiness, here we come!
This is not wired up in the docs just yet.
It is a draft for now.


lfrancke:
Rendered view: https://deploy-preview-418--stackable-docs.netlify.app/home/nightly/policies.html

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+** This means you can upgrade the control plane independently of the user-facing products
+
+* All deprecations follow a predictable lifecycle that allows planning of upgrades ahead of time
+
+The guarantees we give are on multiple levels as detailed below:
+
+* The SDP itself
+* CustomResourceDefinition (CRD) versions
+* Supported product versions
+* Supported OpenShift & Kubernetes
```
soenkeliebau, line 17: The other two points have 'version' attached, is there a reason this one is different?

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+All these policies may evolve as we learn what works best for our customers and for us.
+
+These policies are from *June 2023*.
+
+
+== Stackable Data Platform Lifecycle Policy
+
+NOTE: This policy concerns releases of our platform as a whole and how long and to which extent we support each version.
+
+We do releases of our Stackable Data Platform.
```
soenkeliebau, line 28: Bit of a stub sentence .. can this be merged with the following paragraph?

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+These releases get a name based on the date they have been released (e.g. `23.4`, `23.7`). This name does not follow Semantic Versioning (Semver). We may release patches for a release, which then follow the PATCH naming semantics of Semver (e.g. `23.4.1`). See below for our policy on patches for the SDP.
```
soenkeliebau, line 30:
```suggestion
These releases get a name based on the year and month they have been released in (e.g. `23.4`, `23.7`). This name does not follow Semantic Versioning (Semver). We may release patches for a release, which then follow the PATCH naming semantics of Semver (e.g. `23.4.1`). See below for our policy on patches for the SDP.
```

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+=== Full Support Phase
+
+This phase begins once a new non-patch SDP version is released and ends after a 6-month period OR 3 months after the general availability of the superseding non-patch release, whichever is later.
+
+The following pictures show both scenarios:
+
+TODO: Insert diagrams here
+
+We _will_ release new patch releases of a SDP release (e.g. `23.4.1`) for any issues we deem Critical or High (see definition below) once fixes become available.
```
soenkeliebau, line 46: Maybe add "or for any other reason we see fit" ?

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+IMPORTANT: As of June 2023 all our CRDs are versioned as `alpha1`. We will start introducing other versions later in 2023.
+
+`CustomResourceDefinition`s at Stackable are versioned.
+
+Our policies around CRD versioning are inspired by the https://kubernetes.io/docs/reference/using-api/deprecation-policy/[Kubernetes Deprecation Policy].
+
+Specifically we try to follow these rules:
+
+* API elements may only be removed by incrementing the version of the API group.
```
soenkeliebau, line 73: These three points are too technical and seem to require a fair bit of know how about k8s resource versioning.

At the least I'd suggest an "introductory" paragraph to guide people towards the content here.. or maybe abstract away from the technical details a bit .. I'm not completely sure myself.

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+* API lifetime is determined by the API stability level
+** Alpha API versions may be removed in any release without prior deprecation notice.
+
+** Beta API versions are deprecated at a minimum of 1 platform releases after introduction and removed at a minimum of 1 platform releases after deprecation.
+
+** GA (stable) API versions are deprecated at a minimum of 2 platform releases after introduction and removed at a minimum of 1 platform releases after deprecation.
+
+Similar to the Exception noted by the Kubernetes project itself, we will also evolve these policies as we go along and find that one of the rules doesn't fit a situation.
+
+NOTE: According to these rules it is legal to _add_ fields to a CRD without increasing the version as long as there is a default for this field.
```
soenkeliebau, line 88: I think instead of specifying this one concrete rule I'd rather have a paragraph about api stability in general and how it is okay to change stuff, as long as nothing breaks..

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+
+== Product Lifecycle Policy
+
+SDP ships with a lot of different downstream products (e.g. Apache HBase, Apache Superset).
+
+All of these products follow their own version semantics and release cadences and lifecycles, which is why we do not define a product lifecycle policy based on version numbers alone.
+
+For every one of the products we ship we always support one LTS (Long Term Support) release line, which we generally recommend to use. A release line usually means that we are going to keep a `major.minor` release stable but will include newer patch versions in later SDP releases.
```
soenkeliebau, line 97: "define"  ?

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+Some products (e.g. Trino) don't follow Semver rules, for those we will follow separate rules and clearly document what version is considered LTS.
+
+Every LTS release line is shipped for at least one full year.
+After a year we may switch to a new release line.
```
soenkeliebau, line 102:
```suggestion
After a year we may switch to a new release line - but there will always be at least an overlap of one release in which the old LTS version is deprecated, but a new LTS version is available.
```

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+We do honor the same deprecation policy for non-LTS products as for LTS products, but we do not guarantee a long term support for these versions. They may be deprecated faster.
+
+=== Deprecation
+
+Every product version that gets removed will be deprecated for at least 1 SDP release before removal.
+This guarantees that users can update the operators (e.g. from 23.1 to 23.4) without the need to simultaneously update the product version as well. The flow is to first update the control plane (the operators) and afterward the product versions if desired (e.g. when the currently used version is now deprecated).
+
+=== Definition of "support"
+
+We will ship new versions of the LTS release line in our currently supported SDP releases (see above) for any issues we deem Critical or High in severity when they become available.
```
soenkeliebau, line 116: Mention non-semver products?
eg. Trino .. not even sure how that'd work there ..

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+It is our explicit goal to limit the amount of times we have to ship a version of the products that deviates from the original upstream source.
+
+We may ship new versions for existing SDP releases for other issues as well.
+
+
+== OpenShift & Kubernetes Support Policy
+
+For every SDP release we will publish a list of supported Kubernetes versions.
+
+We are aiming to support the last three Kubernetes versions but will make case-by-case decisions by taking into account the currently supported OpenShift versions as published by RedHat where it is our goal to support all versions that are in Full or Maintenance support. As the releases may be overlapping we might not always support the latest Kubernetes or OpenShift versions when we release a SDP version.
```
soenkeliebau, line 129: Weird switch from kubernetes versions to openshift versions..

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+
+== Support Policy (Security & Bugs)
+
+Stackable will analyze all published security vulnerabilities (e.g. CVEs but other sources may apply as well) for all the products we support. We take various sources into account when assigning a criticality. Among those sources is the NVD database, but we place higher value on the self-assessments by the projects themselves, and we will additionally evaluate all vulnerabilities in the context of how they are used in the Stackable Data Platform.
```
soenkeliebau, line 134:
```suggestion
Stackable will analyze all published security vulnerabilities (e.g. CVEs but other sources may apply as well) for all the products we support as well components developed by us and their dependencies. We take various sources into account when assigning a criticality. Among those sources is the NVD database, but we place higher value on the self-assessments by the projects themselves, and we will additionally evaluate all vulnerabilities in the context of how they are used in the Stackable Data Platform.
```

soenkeliebau:
Left a few comments all over the place. Nothing significant or compelling, mostly things where I maybe had to read a sentence twice to grok it or similar things ..

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+We _may_ release new patch releases of a SDP release (e.g. `23.4.1`) for any issues we deem Critical or High (see definition below) once fixes become available.
+
+TODO: Insert diagram
+
+=== General
+
+We may release patches for other reasons.
+
+Customers are expected to upgrade their SDP environment to the most current supported patch (`23.4.z`) version.
```
sbernauer, line 60: Something along this lines
```suggestion
Customers are expected to upgrade their SDP environment to the most current supported patch (`23.4.z`) version.
Customers need to go throw every intermediate minor release of the SDP, so an upgrade from `23.1.x` to `23.7.x` would need to upgrade to `23.4.z` previously.
```

modules/ROOT/pages/policies.adoc
lfrancke: I added something similar

modules/ROOT/pages/policies.adoc
lfrancke: I tried something. What do you think?

modules/ROOT/pages/policies.adoc
lfrancke: This is a verbatim copy of the K8s policy. I'm happy to hear suggestions.

modules/ROOT/pages/policies.adoc
```diff
@@ -0,0 +1,148 @@
...
+* API elements may only be removed by incrementing the version of the API group.
+
+* API objects must be able to round-trip between API versions in a given release without information loss, except for whole REST resources that do not exist in some versions.
+
+* An API version in a given track may not be deprecated in favor of a less stable API version.
+
+* API lifetime is determined by the API stability level
+** Alpha API versions may be removed in any release without prior deprecation notice.
+
+** Beta API versions are deprecated at a minimum of 1 platform releases after introduction and removed at a minimum of 1 platform releases after deprecation.
```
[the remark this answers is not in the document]
lfrancke: Suggestions?
Again, verbatim copy from K8s

modules/ROOT/pages/policies.adoc
lfrancke: It would be helpful if you could suggest something here as I couldn't fully follow the discussion.

modules/ROOT/pages/policies.adoc
lfrancke: I'm not sure I understand

modules/ROOT/pages/policies.adoc
lfrancke: Good point, yeah. Me neither. Need to discuss.

modules/ROOT/pages/policies.adoc
lfrancke: Is this better?

modules/ROOT/pages/policies.adoc
sbernauer: Hmm this gives us more flexibilty, but kind of moves the decision into the future. I would prefer to start with this "harsh" policy and maybe relax that in the future.
Maybe we can also use something like

You need to go throw every intermediate minor release of the SDP, so an upgrade from `23.1.x` to `23.7.x` would need to upgrade to `23.4.z` previously, unless the release notes state otherwise.

stackfab (approved):
Found no "breaking change". GTCs don't reference CRD policies (yet), so this should go well.

soenkeliebau:
I've looked over the comments and think I am happy to merge this as is for now. 
We can always refine later after getting some feedback from customers.

PR #426, opened 2023-07-21
Document empty string for configOverride

sbernauer:
For https://github.com/orgs/stackabletech/discussions/30


sbernauer:
@dmasice would you consider this helpful information or is something missing?

dmasice:
@sbernauer Looks good to me, but maybe reference it too from the "Configuration & Environment Overrides" in the operator docs?

maltesander (approved):
LGTM!

sbernauer:
@dmasice linking needs to be done in every operator individually, so we can mgere this change and add the linking afterards

PR #432, opened 2023-07-27
simplify and fix getting-started page in main docs

### LINKED ISSUE #428: Getting Started uses more complicated kind invocation than need be
The docs currently say

```
kind create cluster --name quickstart --config - << EOF
---
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
- role: worker
  kubeadmConfigPatches:
    - |
      kind: JoinConfiguration
      nodeRegistration:
        kubeletExtraArgs:
          node-labels: node=quickstart-1
```

but all of that "extra" stuff is not really needed.
It is however used further down the page:

```kubectl get nodes --selector=node=quickstart-1 -o wide```

This ticket should ideally get rid of the "extra" bit and find another way for the kubectl command to work.
If that is not possible I'd like to have a section above this that says that you only need the extra bits for the purpose of this tutorial but for a real one use "this" command instead.

adwk67:
fixes #428 

N.B. should be backported to the `release/23.7` branch and docs rebuilt.


lfrancke (approved):
Thank you!
Do we need a checkbox in our release script that we update this page?

adwk67:
Yes, I think we do. I'll add it to 23.11 and also to the template.

PR #435, opened 2023-08-04
Add cockpit docs

nightkr:
I've probably screwed this up pretty bad.. the header bar behaves very strangely for me locally compared to on docs.stackable.tech.

Ideally this would also only show up for nightly docs (currently), not sure what the ideal way to do that is.

### DIFF
--- a/modules/ROOT/nav2.adoc
+++ b/modules/ROOT/nav2.adoc
@@ -1,3 +1,9 @@
 // This second nav file is for another menu entry at the end of the navigation menu
 // see: https://docs.antora.org/antora/3.0/navigation/organize-files/#multiple-files-per-module
+* Experimental
+** xref:cockpit:index.adoc[]
++
+--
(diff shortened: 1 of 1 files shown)


fhennig:
The changes look fine to me in the files, I can take a look at it more why it might look wonky if you want to!

The navbar is part of the template, and so far the template is always the same for all versions. I don't think there is a way to make this only show up in certain versions.

fhennig:
Okay, I noticed a couple of things. The two playbook files still reference different branches (In one of them it still says FIXME). 

I think the antora.yml should not be part of `home` but should be more like the one for [stackablectl](https://github.com/stackabletech/stackablectl/blob/main/docs/antora.yml). We have the `home` component for all the operators and the main docs, which are all versioned together. But the stackablectl docs are not versioned, and are in a separate component. I think the cockpit docs would also be unversioned, right? That should also fix the 404 issue I believe.

PR #442, opened 2023-08-24
ADR on Discovery 2.0

sbernauer:


fhennig:
@nightkr What do you mean with "I would disagree with the entire premise of this." What's the premise?

I also read it now, and wanted to share some thoughts too. I think the ease of use of the ConfigMap is actually really nice. I think most discovery configmaps contain connect strings or files, and those can then either be mounted as an env var or as files, so it's very easy to use.

The document mentions that cross-namespace is an issue with ConfigMaps, and also name clashes. If these are problems, I'd like to see these aspects mentioned as decision drivers in the earlier section of the ADR. I personally do not think that these two things are very important.

nightkr:
> What's the premise?

The primary driver for this (motivations 2/3) seem to be authentication. I disagree that discovery is the right place to tackle either.

> The document mentions that cross-namespace is an issue with ConfigMaps, and also name clashes. If these are problems, I'd like to see these aspects mentioned as decision drivers in the earlier section of the ADR. I personally do not think that these two things are very important.

Agreed that both of these are effectively treated as decision drivers looking at the rest of the document. I'd also disagree with the document about discovery being the right place to tackle either of those either.

Cross-namespace use: Kubernetes namespaces are effectively a tenancy border, we should be *very* careful blurring that line. To me, cross-namespace use should be something you request explicitly, and which is ideally isolated from other "tenants" (`ZookeeperZnode` ticks both of these boxes, but isolation is easier done for a storage system like ZK or HDFS than something like Trino).

fhennig:
Hmm, can you elaborate why the discovery mechanism shouldn't also contain information on the authentication mechanism? It makes sense to me, it's information that a client would need when connecting to the service.

maltesander:
> Cross-namespace use: Kubernetes namespaces are effectively a tenancy border, we should be _very_ careful blurring that line. To me, cross-namespace use should be something you request explicitly, and which is ideally isolated from other "tenants" (`ZookeeperZnode` ticks both of these boxes, but isolation is easier done for a storage system like ZK or HDFS than something like Trino).

The discovery-operator proposal is partly based on the ZNode logic. E.g. for deploying / mounting configmaps across namespaces.

soenkeliebau:
It sounds to me like this is worth an online debate?
Would everybody be happy for me to organize and prepare a meeting to discuss?

PR #460, opened 2023-09-26
OIDC AuthenticationClass ADR

fhennig:
related: 

- this demo of OIDC connectivity with a couple of products: https://github.com/stackabletech/stackablectl/pull/283
- OpenID Connect epic: https://github.com/stackabletech/issues/issues/431


sbernauer:
Meeting minutes from 2023-10-23
```
# Non-breaking switch LDAP to be consistent
apiVersion: authentication.stackable.tech/v1alpha1
kind: AuthenticationClass
metadata:
  name: ldap-simple
spec:
  provider:
    ldap:
      hostname: my.ldap.server 
      port: 389 
      searchBase: ou=users,dc=example,dc=org 
      bindCredentials:
        secretClass: openldap-bind-credentials
# OR
      reference: my-ldap-connection-details
      bindCredentials:
        secretClass: openldap-bind-credentials

# New OIDC provider
apiVersion: authentication.stackable.tech/v1alpha1
kind: AuthenticationClass
metadata:
  name: keycloak-oidc
spec:
  provider:
    oidc:
      reference: my-keycloak-connection-details
      bindCredentials: # mandatory
        secretClass: keycloak-trino-client
# OR (order matters in Rust enum code!)
      hostname: keycloak.mycorp
      port: # optional, default depends on wether tls is enabled, 80 or 443
      root: / # mandatory, defaults to "/" (effective optional). We will append "/.well-known/openid-configuration" internally when needed by the product.

      # usernameClaim is not part of this struct, "sub" will always be used to extract the user identifier that is used within OPA rules.
      # Additionally we document (TBD recommend???) users to configure their OIDC software to send the actual username instead of the uuid in the sub field
      # in case they use old-school products that assume a user only has a username - and nothing more (e.g. hdfs, kafka).

      tls: # optional commons tls config, which is an common struct from operator-rs. Implementation can be copied from LDAP.
        verification:
          none: {}
          # OR
          server:
            caCert:
              webPki: {}
              # OR
              secretClass: tls # Reference to SecretClass below
      clientCredentials: # mandatory
        secretClass: keycloak-trino-client
```

sbernauer:
Current state:
```
# Leave as is
apiVersion: authentication.stackable.tech/v1alpha1
kind: AuthenticationClass
metadata:
  name: ldap-simple
spec:
  provider:
    ldap:
      hostname: my.ldap.server 
      port: 389 
      searchBase: ou=users,dc=example,dc=org 
      bindCredentials:
        secretClass: openldap-bind-credentials
      tls:
        verification: none

# New OIDC provider
apiVersion: authentication.stackable.tech/v1alpha1
kind: AuthenticationClass
metadata:
  name: keycloak-oidc
spec:
  provider:
    oidc:
      hostname: keycloak.mycorp
      port: # optional, default depends on wether tls is enabled, 80 or 443
      root: / # mandatory, defaults to "/" (effective optional). We will append "/.well-known/openid-configuration" internally when needed by the product.

      # usernameClaim is not part of this struct, "sub" will always be used to extract the user identifier that is used within OPA rules.
      # Additionally we document (TBD recommend???) users to configure their OIDC software to send the actual username instead of the uuid in the sub field
      # in case they use old-school products that assume a user only has a username - and nothing more (e.g. hdfs, kafka).

      tls: # optional commons tls config, which is an common struct from operator-rs. Implementation can be copied from LDAP.
        verification:
          none: {}
          # OR
          server:
            caCert:
              webPki: {}
              # OR
              secretClass: tls # Reference to SecretClass below
      # No(!) oidc client credentials, these needs to be specified on the Product CR using the AuthClass

# Usage in product
apiVersion: trino.stackable.tech/v1alpha1
kind: TrinoCluster
metadata:
  name: trino
spec:
  clusterConfig:
    authentication:
      - authenticationClass: static-trino-users
      - authenticationClass: keycloak-oidc
        oidcClientCredentialsSecret: keycloak-trino-client # optional, mandatory in case spec.provider of the referenced AuthClass is oidc
                                                           # and the product does not support PKCE or ImplicitFlow authentication mode against Keycloak.
      # Long term we want to do the same for LDAP
      # This is breaking, as users need to specify the bind credentials here
      - authenticationClass: my-openldap
        ldapBindCredentialsSecret: my-openldap-trino-bind # optional (as LDAP does not always require a bind user)
```

sbernauer:
Meeting minutes:
```
apiVersion: authentication.stackable.tech/v1alpha1
kind: AuthenticationClass
metadata:
  name: keycloak
spec:
  provider:
    oidc:
      hostname: "$KEYCLOAK_HOSTNAME"
      port: $KEYCLOAK_PORT
      rootPath: /realms/master
      scopes: [openid]
      # docs:
      # This is a hint for the products on a best-effort base, most of the products will ignore this value.
      # E.g. Superset Superset uses this to configure the correct claims to extract from the userinfo endpoint as well as a nice icon
      # => We only start with Keycloak and Custom in v1, but might extend it in the future
      provider: Keycloak / Google / Github / Azure / Openshift / Okta / Custom # mandatory enum, defaults to Custom
      tls:
        verification:
          none: {}
---
apiVersion: trino.stackable.tech/v1alpha1
kind: TrinoCluster
metadata:
  name: trino
spec:
  image:
    productVersion: "414"
    stackableVersion: 23.7.0
  clusterConfig:
    listenerClass: external-unstable
    tls:
      serverSecretClass: tls
    catalogLabelSelector:
      matchLabels:
        trino: trino
    authentication:
      - authenticationClass: keycloak
### CHOSEN
        oidc: # Option<>
          clientCredentialsSecret: trino-keycloak-client # mandatory (% anonymous client connection), error when AuthClass provider is oidc and this is not set
          extraPath: /.well-know/foo # FIELD IS ONLY FOR SUPERSET: optional and we infer it based upon the spec.provider.oidc.provider of the AuthClass.
                                     # We error out in case AuthClass provider is oidc and oidc.provider is Custom and and this is not set
        ldap: # Option<>
          bindCredentialsSecretClass: trino-openldap-bind # optional for some products, mandatory for others
### CHOSEN


        # OR
        oidcClientCredentialsSecret: trino-keycloak-client # mandatory (% anonymous client connection)
        oidcExtraPath: /.well/know/foo # mandatory

        # OR
      - oidc:
          authenticationClass: keycloak # snafu error when types do not match
          clientCredentialsSecret: trino-keycloak-client
          extraPath: /.well/know/foo

        # OR (discarded)
        secret:
          oidc: trino-keycloak-client
        oidcExtraPath: /.well/know/foo

        # OR (discarded)
        secret: trino-keycloak-client
        oidcExtraPath: /.well/know/foo
        # Might cause problems when users switch from LDAP to OIDC
        /stackable/secret/trino-keycloak-client/bindUser
        # vs
        /stackable/secret/trino-keycloak-client/clientId

        # In the future
        ldapBindUserSecret: trino-openldap-bind
    authorization:
      opa:
        configMapName: opa
        package: trino
  coordinators:
    roleGroups:
      default:
        replicas: 1
  workers:
    roleGroups:
      default:
        replicas: 1
---
apiVersion: v1
kind: Secret
metadata:
  name: trino-keycloak-client
stringData:
  clientId: trino
  clientSecret: "{{ keycloakTrinoClientSecret }}"
```

maltesander:
I tried to implement this on trino but could not get kube-rs to generate the correct CRD when flattening an optional enum (its always mandatory). Dont mind the names, just an example.
```
#[derive(Clone, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
#[serde(rename_all = "camelCase")]
pub struct TrinoAuthenticationClassRef {
    pub authentication_class: String,
    #[serde(flatten)]
    pub secret: Option<TrinoAuthenticationSecret>,
}

#[derive(Clone, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
#[serde(rename_all = "camelCase")]
pub enum TrinoAuthenticationSecret {
    Oidc {
        #[serde(rename = "clientCredentialsSecret")]
        client_credentials_secret: String,
    },
}
```
generates
```
                clusterConfig:
                  description: Trino cluster configuration options.
                  properties:
                    authentication:
                      default: []
                      description: Authentication options for Trino.
                      items:
                        oneOf:
                          - required:
                              - oidc
                        properties:
                          authenticationClass:
                            type: string
                          oidc:
                            properties:
                              clientCredentialsSecret:
                                type: string
                            required:
                              - clientCredentialsSecret
                            type: object
                        required:
                          - authenticationClass
                        type: object
                      type: array
```
Any ideas?

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+As mentioned before, we expect the user to already operate the identity provider.
+
+== Decision Drivers
+
+* DRY - don't repeat yourself: information should ideally only be configured in one spot.
+* flexible: Different variants of client and product instance mapping should be supported.
+* comprehensible: It not require too much documentation to explain how a product configuration is assembled based on the manifests that the user deploys.
+* comprehensive: supporting all our existing products (if they support ODIC) as well as all (or most) OIDC providers listed above.
+* principle of least surprise: OIDC should be configured in a similar way to LDAP.
```
NickLarsenNZ, line 116: We just need to clarify what we mean by _similar way to LDAP_.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+      serverSecretClass: tls
+    catalogLabelSelector:
+      matchLabels:
+        trino: trino
+    authentication:
+      - authenticationClass: keycloak
+### CHOSEN
+        oidc: # Option<>
+          clientCredentialsSecret: trino-keycloak-client # mandatory (% anonymous client connection), error when AuthClass provider is oidc and this is not set
+          extraPath: /.well-know/foo # FIELD IS ONLY FOR SUPERSET: optional and we infer it based upon the spec.provider.oidc.provider of the AuthClass.
```
NickLarsenNZ, line 164: We should remove the `extraPath` config and hard code `/.well-known/openid-configuration` in the operator (eg: superset) for products that don't support it natively.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+    oidc:
+      hostname: "$KEYCLOAK_HOSTNAME"
+      port: $KEYCLOAK_PORT
+      rootPath: /realms/master
+      scopes: [openid]
+      # docs:
+      # This is a hint for the products on a best-effort base, most of the products will ignore this value.
+      # E.g. Superset Superset uses this to configure the correct claims to extract from the userinfo endpoint as well as a nice icon
+      # => We only start with Keycloak and Custom in v1, but might extend it in the future
+      provider: Keycloak / Google / Github / Azure / Openshift / Okta / Custom # mandatory enum, defaults to Custom
```
NickLarsenNZ, line 139: `Custom` sounds ike "bring your own implementation", whereas I think the intention of this is to just use an OIDC compliant provider.
Suggestions for rename (ignore case):
- `Generic`
- `Compliant`
- `Default`

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+----
+apiVersion: authentication.stackable.tech/v1alpha1
+kind: AuthenticationClass
+metadata:
+  name: keycloak
+spec:
+  provider:
+    oidc:
+      hostname: "$KEYCLOAK_HOSTNAME"
+      port: $KEYCLOAK_PORT
```
NickLarsenNZ, line 132: We discussed whether this had to be split or could be combined. Trino needs this split, so we will keep hostname and port as separate options too (instead of writing code to parse them).

NickLarsenNZ: Also, port should default to `None`. The TLS config will determine whether the scheme will be `https://` or `http://` using implied ports (unless overridden)

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+      rootPath: /realms/master
+      scopes: [openid]
+      # docs:
+      # This is a hint for the products on a best-effort base, most of the products will ignore this value.
+      # E.g. Superset Superset uses this to configure the correct claims to extract from the userinfo endpoint as well as a nice icon
+      # => We only start with Keycloak and Custom in v1, but might extend it in the future
+      provider: Keycloak / Google / Github / Azure / Openshift / Okta / Custom # mandatory enum, defaults to Custom
+      tls:
+        verification:
+          none: {}
```
NickLarsenNZ, line 142: TLS should be on, and verified by default, but opt-out for those times where you just need to test something.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
```diff
@@ -0,0 +1,221 @@
...
+kind: AuthenticationClass
+metadata:
+  name: keycloak
+spec:
+  provider:
+    oidc:
+      hostname: "$KEYCLOAK_HOSTNAME"
+      port: $KEYCLOAK_PORT
+      rootPath: /realms/master
+      scopes: [openid]
```
NickLarsenNZ, line 134: We might also want to add in `profile` and `email` (and any scopes for group/role mappings?) like other apps seem to use?

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
NickLarsenNZ: We also talked about renaming the key to `providerHints` or `providerQuirks`

NickLarsenNZ: Someone (@Techassi or @fhennig?) suggested:
```
providerHint: x # Option<ProviderHint>
```

NickLarsenNZ: Superset and Airflow need this hint. So if unset, they won't work.
**Decision**: put is in the docs that it is mandatory if using those products.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
Techassi: This is completely up to the user/admin. We can however mention the recommended set of scopes in the docs, which are: `openid`, `profile` and `email`.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
Techassi: Also mention that `None` (so not specifying any hint) basically means we assume the product works with OIDC straight out-of-the-box.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
Techassi: Removed in a5471f6.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
Techassi: Yes, the ADR needs some textual adjustments.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
fhennig: _Technically_ the recommended scopes depends on the products you're connecting. I.e. druid only wants openid, but Superset uses the profile and email. 

I think there's almost no downside to requesting a scope, so maybe we could default to request all 3 scopes?

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
fhennig: i'd really like to stick to the TLS struct we have for the scope of this epic, and spin this off into its own task to be tackled afterwards. I can volunteer to create such a ticket

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
fhennig: :thumbsup:

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
fhennig: I have clarified it in this commit: https://github.com/stackabletech/documentation/pull/460/commits/ff23516e9a79db4afc84c929f6eb1046b1e725c9

Is that better?

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
fhennig: I thought Superset also had a sort of "default" provider? 

Anyways, I'm in support of the Option<ProviderHint>

NickLarsenNZ: Yeah @dervoeti mentioned that Superset needs to have the hint (for both building URLs and to select one of a handful of options it supports, as well as for the logo to display).

_There is a chance that information is outdated as there were other casual discussions during the day._

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
NickLarsenNZ: Is this ok @nightkr?

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
NickLarsenNZ: > We have previously designed the AuthenticationClass mechanism and how it integrates into product CRDs, at the time to support LDAP authentication. The OIDC configuration should have similar ergonomics so users are not surprised and we get a coherent platform.

Cool, this adds the context that I was missing. I could suggest an alternativeversion:

_principle of least surprise: The `AuthenticationClass` mechanism was originally designed to support LDAP. Extending this to support OIDC should have similar ergonomics and be obvious to implement._

_I kept the original prefix to be consistent with the other items above. I think perhaps they should begin with a capital letter_.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
nightkr: Aight.

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -150,33 +173,15 @@ spec:
...
-  coordinators:
-    roleGroups:
-      default:
-        replicas: 1
-  workers:
-    roleGroups:
-      default:
-        replicas: 1
+        oidc:
+          # A reference to the OIDC client credentials secret, which consists of a client id and secret
```
NickLarsenNZ, line 180: I think we should include the exact keys that would be expected in the secret (eg: `client_id` and `client_secret`). I know they'll be documented elsewhere, but will be useful for any reader of the comment.

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -150,33 +173,15 @@ spec:
...
+          clientCredentialsSecret: trino-keycloak-client
+
+          # Additional scopes required for this specific product. It will get merged with the above configured scopes
```
NickLarsenNZ, line 183: Same as my other comment about scopes, product specific scopes are likely built into the code and are not user configurable.

Do we know of any products that can make use of user-defined scopes?

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -187,6 +192,38 @@ stringData:
...
+metadata:
+  name: trino
+spec:
+  image:
+    productVersion: "414"
+    stackableVersion: 23.7.0
+  clusterConfig:
+    # Other required config options omitted for brevity
+    authentication:
+      - authenticationClass: open-ldap
```
NickLarsenNZ, line 213: I think this is going from a list to an enum. A list would suggests multiple methods can be configured at the same time.

NickLarsenNZ (commented):
Thanks for the ADR updates @Techassi.
I have noted some things down that came to mind.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
NickLarsenNZ: I'll mark this as resolved. My latest review opens a question about the need for user defined scopes (as opposed to hard-coded scopes as per each products requirements).

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
NickLarsenNZ: There is still a question of whether `hostname` and `port` should be split, or a url is provided:

- split:
  - easier to pass on to products which expect them to be split
  - http scheme is determined by whether TLS is enabled or not
- combined:
  - less settings
  - no need to reconstruct the URL (eg: for when testing things with cURL)
  - removes the need for the rootPath, as that would be part of the url

_@nightkr I hope I've correctly captured what was discussed today in casual conversation_

We decided to keep building as if they are split, but we can always change over closer to merge time.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
NickLarsenNZ: The latest changes resolve this

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -72,7 +78,10 @@ druid.auth.pac4j.oidc.discoveryURI: '${env:KEYCLOAK_DISCOVERY_URL}'
 
 ==== Superset
 
-Does not support reading an OIDC discovery URL but needs: API base URL, auth URL, token URL. Also the Superset client wants the _email_, _profile_ and _oidc_ scope. We _could_ also implement fab-oidc support for Superset, so it has proper OIDC support. Needs to be maintained though.
+Does not support reading an OIDC discovery URL but needs: API base URL, auth URL, token URL.
```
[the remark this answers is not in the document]
Techassi: 118145b

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -72,7 +78,10 @@ druid.auth.pac4j.oidc.discoveryURI: '${env:KEYCLOAK_DISCOVERY_URL}'
...
+Also the Superset client wants the _email_, _profile_ and _oidc_ scope.
```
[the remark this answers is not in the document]
Techassi: 118145b

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -128,19 +141,29 @@ metadata:
 spec:
   provider:
     oidc:
+      # Hostname of the IdP. Like "idp.mycompany.corp"
       hostname: "$KEYCLOAK_HOSTNAME"
+
+      # Optional port number to use. If unspecified, this will default to 80 if TLS is disabled, 443 otherwise.
```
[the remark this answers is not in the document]
Techassi: 8e5cca1

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
Techassi: b8d3813

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
Techassi: Which is supported for some products. Some products might not be able to configure multiple authentication methods, for example Superset. In that case, we just error out if `classes.length() > 1`.

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -36,23 +41,24 @@ It is best practice to have one client per connecting product.
 This allows the user to define exactly from which hosts the client can be used, as well as valid redirect URLs.
 Sometimes a separate client per product is also technically required, because a product might require specific configuration for a particular product, i.e. rewriting claims.
 
-This leaves multiple different mappings from products and product instances to configured clients; making these use cases easy to configure is an important consideration.
+Configuring multiple different mappings from products and product instances needs to take place at the client level; making these use cases easy to configure is an important consideration.
```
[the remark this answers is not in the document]
Techassi: 6037b10

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
Techassi: Trino has support for it. Other products might also support it. See `http-server.authentication.oauth2.scopes` at https://trino.io/docs/current/security/oauth2.html#trino-server-configuration

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
```diff
@@ -128,19 +141,29 @@ metadata:
...
+      # Optional root path appended to the hostname. This defaults to "/".
       rootPath: /realms/master
-      scopes: [ openid, email, profile ] # User can specify whatever scopes they want, these 3 are recommended for OIDC
-      # docs:
-      # This is a hint for the products on a best-effort base, most of the products will ignore this value.
-      # E.g. Superset Superset uses this to configure the correct claims to extract from the userinfo endpoint as well as a nice icon
-      # => We only start with Keycloak in v1, but might extend it in the future
-      providerHint: Keycloak # Option<ProviderHint>, None means no hint, just use OIDC as is
+
+      # User can specify whatever scopes they want, these three are recommended for OIDC.
```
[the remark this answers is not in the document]
Techassi: da81fcd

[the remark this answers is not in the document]
Techassi: I guess we should decide on it rather quickly. If we decide we don't see an immediate value in having support for user defined scopes, we might opt to remove them from version 1 of our AuthenticationClass.

Adding it in a later point in time is possible without introducing any breaking change.

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
NickLarsenNZ: Ah yes, I understand now :+1:

modules/contributor/pages/adr/ADR032-oauth-oidc-support.adoc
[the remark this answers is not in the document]
Techassi: The scopes are needed. Ultimately the IdP has control over which data is returned based on the requested scopes. Most IdPs however follow conventions. Providing these scopes can help to "patch" any missing data.

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
Techassi: I think we could handle the URL as a single value. Operators which need to set parts of the URL separately can do so by extracting required data from `url::Url` provided by `operator-rs` via the associated function `endpoint_url`.

modules/contributor/pages/adr/ADR032-oidc-support.adoc
```diff
@@ -0,0 +1,245 @@
...
+* Date: 2023-11-14
+
+Technical Story: https://github.com/stackabletech/issues/issues/431
+
+== Problem statement
+
+OIDC is a widespread authentication mechanism, supported by most modern data products as well as identity providers.
+OIDC is also used inside the Gaia-X ecosystem.
+Keycloak is also widespread as a self-hosted, internal identity provider.
+Just as with LDAP, we expect users to already have an identiy provider running; deploying an identity provider is out of scope of the design.
```
adwk67, line 25:
```suggestion
Just as with LDAP, we expect users to already have an identity provider running; deploying an identity provider is out of scope of the design.
```

modules/contributor/pages/adr/ADR032-oidc-support.adoc
```diff
@@ -0,0 +1,245 @@
...
+
+Problems include the slightly different ways of configuring products; deciding on how OIDC clients are configured and how to provide client credentials.
+
+== Context
+
+Useful information to consider in the decision.
+
+=== OIDC clients and product-client-mappings
+
+To configure a product to use ODIC, it needs information about where to find the OIDC provider (typically the URL of the discovery endpoint), as well as client credentials to authenticate with the provider.
```
adwk67, line 37:
```suggestion
To configure a product to use OIDC, it needs information about where to find the OIDC provider (typically the URL of the discovery endpoint), as well as client credentials to authenticate with the provider.
```

modules/contributor/pages/adr/ADR032-oidc-support.adoc
```diff
@@ -0,0 +1,245 @@
...
+The connection information is generic and shared between all clients, but the crendentials are product specific.
+
+It is best practice to have one client per connecting product.
+This allows the user to define exactly from which hosts the client can be used, as well as valid redirect URLs.
+Sometimes a separate client per product is also technically required, because a product might require specific configuration for a particular product, i.e. rewriting claims.
+
+Configuring multiple different mappings from products and product instances needs to take place at the client level; making these use cases easy to configure is an important consideration.
+These mapping can be configuration option mapping or even JWT claim mapping.
```
adwk67, line 45:
```suggestion
These mappings can be configuration option mappings or even JWT claim mappings.
```

modules/contributor/pages/adr/ADR032-oidc-support.adoc
```diff
@@ -0,0 +1,245 @@
...
+* Dex
+* Okta
+
+As mentioned before, we expect the user to already operate the identity provider.
+
+== Decision drivers
+
+* Don't repeat yourself: Information should ideally only be configured in one spot.
+* Flexible: Different variants of client and product instance mappings should be supported.
+* Comprehensible: Users should not be overwhelmed by complicated documentation. After seeting up one product, users should be able to fairly easily setup other products as well.
```
adwk67, line 126:
```suggestion
* Comprehensible: Users should not be overwhelmed by complicated documentation. After setting up one product, users should be able to fairly easily setup other products as well.
```

adwk67 (commented):
Just some minor typos.

fhennig:
Thanks @adwk67 !

razvan (approved):
lgtm

modules/contributor/pages/adr/ADR032-oidc-support.adoc
```diff
@@ -0,0 +1,245 @@
+= ADR0XX: OIDC Support
```
NickLarsenNZ, line 1: The filename is numbered, but the title isn't

modules/contributor/pages/adr/ADR032-oidc-support.adoc
```diff
@@ -0,0 +1,245 @@
...
+      # Optional root path appended to the hostname. This defaults
+      # to "/".
+      rootPath: /realms/master
+
+      # User configurable scopes depending on the product requirements.
+      # The following three scopes are usually required, and might even
+      # be hard-coded in the product. Scopes defined here are probably
+      # merged by the product into any default scopes it sets
+      # out-of-the-box.
+      scopes: [ openid, email, profile ]
```
NickLarsenNZ, line 164: If I recall correctly, Natalie was saying these scopes are relevant to the iDP rather the product that we run.

NickLarsenNZ (approved):
lgtm

modules/contributor/pages/adr/ADR032-oidc-support.adoc
NickLarsenNZ: I've reworded this in a later commit. @Techassi, you can resolve this if you're happy with the edit.

modules/contributor/pages/adr/ADR032-oidc-support.adoc
NickLarsenNZ: I've set this

modules/contributor/pages/adr/ADR032-oidc-support.adoc
fhennig: The IDP defines the scopes that are available, but the product decides what it needs. I.e. the product might need an email adress. in IDP A this might be accessible by requiesting the "profile" scope, in IDP B it might be a specific "email" scope. That's how I understand it works.

NickLarsenNZ: That _was_ also my understanding, hence my earlier comments (that were marked as resolved) about why it would be user defined if the product has already chosen?

So now I'm confused:
- is it something the user can choose? (Why?)
- is it something already decided by the product (so then why would it be configurable?)
- are there additional scopes the iDP might require you to ask for? (Which suits the current solution, except that we should then call it `additionalScopes` and hard code the known scopes into the product operator).

_Apologies for the terse response, I'm on mobile_

NickLarsenNZ: Ohh, I think I understand what you mean...

The product requires email address, expecting it to be in the returned `email` claim.
- the product leaves it to the user to define which scope their idP requires in the request for that claim to be added.
  - idP A: adds the `email` claim if the `profile` scope is requested
  - idP B: adds the `email` claim if the `email` scope is requested
  - idP C: adds the `email` claim if either scope exists in the request

NickLarsenNZ: We can resolve it once I/we revert that commit

modules/contributor/pages/adr/ADR030-oauth-oidc-support.adoc
Techassi: What's the final consensus here?

NickLarsenNZ: > Operators which need to set parts of the URL separately can do so by extracting required data from url::Url provided by operator-rs via the associated function endpoint_url.

Makes sense, let's go with a single field, `url`.

NickLarsenNZ (approved):
Good to go

PR #461, opened 2023-09-27
concepts: Document caveats with HPAs and PDBs 

sbernauer:
Part of https://github.com/stackabletech/issues/issues/435


modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -15,6 +15,27 @@ The defaults depend on the individual product and can be found below the "Operat
...
+there is no single point of failure (as we have at least 5 nodes available), but also we only have a single spare node left. The reason why you did choose 7 instead of 5
+Zookeeper replicas might be, that you always want at least 2 spares. So by increasing the number of allowed disruptions when you increase the number of replicas probably
+is not what you are trying to achieve when you increase the replicas to increase availability.
+
+=== Multiple replicas to increase performance
+
+For these roles (e.g. HDFS datanodes, HBase regionservers or Trino workers) we allow more than a single Pod to be unavailable, as otherwise rolling re-deployments could take very long.
+
+IMPORTANT: The operators calculate the number of Pods for a give role by adding the number of replicas of every rolegroup that is part of that role.
+In case their are no replicas defined on a rolegroup, one Pod will be assumed for this rolegroup, as the created Kubernetes objects
```
maltesander, line 32:
```suggestion
In case there are no replicas defined on a rolegroup, one Pod will be assumed for this rolegroup, as the created Kubernetes objects
```

modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -15,6 +15,27 @@ The defaults depend on the individual product and can be found below the "Operat
...
+(StatefulSets or Deployments) will default to a single replica as well. However, in case there are
+https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/[HorizontalPodAutoscaler] in place, the number of replicas of the rolegroup
+can change dynamically. In this case the operators might falsely assume that rolegroups have less Pods than they actually have. This is a pessimistic approach,
```
maltesander, line 35:
```suggestion
can change dynamically. In this case the operators might falsely assume that rolegroups have fewer Pods than they actually have. This is a pessimistic approach,
```

maltesander (changes requested):
I find the pod disruption part very hard to read / understand. I try to come up with an alternative later.

fhennig (dismissed):

modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -15,6 +15,27 @@ The defaults depend on the individual product and can be found below the "Operat
...
+
+For these roles (e.g. HDFS datanodes, HBase regionservers or Trino workers) we allow more than a single Pod to be unavailable, as otherwise rolling re-deployments could take very long.
+
+IMPORTANT: The operators calculate the number of Pods for a give role by adding the number of replicas of every rolegroup that is part of that role.
+In case there are no replicas defined on a rolegroup, one Pod will be assumed for this rolegroup, as the created Kubernetes objects
+(StatefulSets or Deployments) will default to a single replica as well. However, in case there are
+https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/[HorizontalPodAutoscaler] in place, the number of replicas of the rolegroup
+can change dynamically. In this case the operators might falsely assume that rolegroups have fewer Pods than they actually have. This is a pessimistic approach,
+as the number of allowed disruption normally stays the same or even increases when the number of Pods increases. So this should be save, but in some cases
+more Pods *could* have been allowed to be unavailable, so rolling re-deployments can take a bit longer than needed.
```
maltesander, line 37: Tried to rephrase a little and use less "we" and "you" (no strong opinion here).
I think performance does not really fit in here or probably deserves its own section?
```suggestion
In general, product roles are split into the following two categories, which serve as guidelines for the default values we apply:

=== Multiple replicas to increase availability

For these roles (e.g. ZooKeeper servers, HDFS journal + namenodes or HBase masters), only a single Pod is allowed to be unavailable. For example, imagine a cluster with 7 ZooKeeper servers, where 4 servers are required to form a quorum and healthy ensemble. By allowing 2 servers to be unavailable, there is no single point of failure (as there are at least 5 servers available). But there is only a single spare server left. The reason to choose 7 instead of e.g. 5 ZooKeeper servers might be, that there are always at least 2 spare servers. Increasing the number of allowed disruptions and increasing the number of replicas is not improving the general availability.

=== Multiple replicas to increase performance

For these roles (e.g. HDFS datanodes, HBase regionservers or Trino workers), more than a single Pod is allowed to be unavailable. Otherwise, rolling re-deployments may take very long.

IMPORTANT: The operators calculate the number of Pods for a given role by adding the number of replicas of every rolegroup that is part of that role.

In case there are no replicas defined on a rolegroup, one Pod will be assumed for this rolegroup, as the created Kubernetes objects (StatefulSets or Deployments) will default to a single replica as well. However, in case there are https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/[HorizontalPodAutoscaler] in place, the number of replicas of a rolegroup can change dynamically. In this case the operators might falsely assume that rolegroups have fewer Pods than they actually have. This is a pessimistic approach, as the number of allowed disruptions normally stays the same or even increases when the number of Pods increases. This should be safe, but in some cases more Pods *could* have been allowed to be unavailable which may increase the duration of rolling re-deployments.
```

maltesander (changes requested):

modules/concepts/pages/operations/index.adoc
```diff
@@ -27,23 +28,26 @@ WARNING: Graceful shutdown is not implemented for all products yet. Please check
...
-   your Kubernetes nodes and use xref:overrides.adoc#pod-overrides[podOverrides] to add a `toleration` for the taint.
+1. *Compute resources*: You can configure the available resource every product has using xref:concepts:resources.adoc[].
+   The defaults are very restrained, as you should be able to spin up multiple products running on your Laptop.
+2. *Autoscaling*: Although not supported by the platform natively yet, you can use
+   https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale[HorizontalPodAutoscaler] to autoscale the number of Pods running for a given rolegroup dynamically based upon resource usage.
+   To achieve this you need to omit the number of replicas on the rolegroup to be scaled, which in turn results in the created StatefulSet not having any replicas set as well.
+   Afterwards you can deploy a HorizontalPodAutoscaler as usual.
+   Please note that not all product-operators have implemented graceful shutdown, so product might be disturbed during scale down.
+   Later platform versions will support autoscaling natively with sensible defaults and will deploy HorizontalPodAutoscaler objects for you.
```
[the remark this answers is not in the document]
maltesander: If you decided it together im fine with it :)

Techassi (approved):

maltesander (approved):
LGTM!

PR #462, opened 2023-09-28
 Rename clusterOperation flag "reconcilePaused" to "reconciliationPaused"

siegfriedweber:
Rename clusterOperation flag "reconcilePaused" to "reconciliationPaused"
 
 This change should be backported to 23.7 and 23.4.

### DIFF
--- a/modules/concepts/examples/cluster-operations.yaml
+++ b/modules/concepts/examples/cluster-operations.yaml
@@ -5,5 +5,5 @@ metadata:
   name: my-cluster
 spec:
   clusterOperation:
-    reconcilePaused: false # <1>
+    reconciliationPaused: false # <1>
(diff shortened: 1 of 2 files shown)


nightkr:
Is this an actual breaking change or a case of the docs mismatching the implementation?

siegfriedweber:
> Is this an actual breaking change or a case of the docs mismatching the implementation?

The documentation was incorrect. The implementation always used `reconciliationPaused`.

sbernauer (approved):
LGTM as this only fixes the docs. Do you take care of back-porting this?

siegfriedweber:
> LGTM as this only fixes the docs. Do you take care of back-porting this?

backported to 23.7 and 23.4

PR #464, opened 2023-10-02
Feat: versioned menu links

### LINKED ISSUE #364: Make top menu link to current version instead of stable
The top menu always links to stable, but it would be better if it linked to the version of the current page.

There are [some variables available in the template](https://docs.antora.org/antora-ui-default/templates/) but it doesn't resolve "231." to "stable". So I couldn't do it within a couple of minutes of trying. It might be doable in a clever or hacky way.

Edit: At the same time, we should probably have a banner like "you're looking at and old version of the docs" to tell the user to read the latest stuff. And also to show to search engines that the 'stable' link is the important one: https://github.com/stackabletech/documentation-ui/issues/36

Currently all our pages always link to stable and that gives those pages a lot of importance. We should keep that.

fhennig:
fixes #364 

requires documentation-ui feature: https://github.com/stackabletech/documentation-ui/pull/48

### DIFF
--- a/modules/ROOT/pages/quickstart.adoc
+++ b/modules/ROOT/pages/quickstart.adoc
@@ -1,6 +1,6 @@
 = Quickstart
 
-:latest-release: https://github.com/stackabletech/stackable-cockpit/releases/tag/stackablectl-1.0.0-rc1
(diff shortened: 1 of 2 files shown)


sbernauer (approved):
Works perfectly fine, thanks!

sbernauer (approved):

Issue #470, opened 2023-10-18
logging: Add section on different sinks (OpenSearch, HDFS, Elastic)


fhennig:
can be just a few sentences

https://docs.stackable.tech/home/stable/concepts/logging

PR #471, opened 2023-10-18
logging docs: mention other sinks

fhennig:


modules/concepts/pages/logging.adoc
```diff
@@ -88,3 +120,4 @@ logging:
 == Further Reading
 
 To get some hands on experience and see logging in action, try out the xref:demos:logging.adoc[logging demo].
+You can read more about the https://vector.dev/docs/setup/deployment/topologies/#centralized[Vector topology] or https://vector.dev/docs/reference/configuration/sinks/[sinks].
```
siegfriedweber, line 123:
```suggestion
The Vector documentation conains more information about the https://vector.dev/docs/setup/deployment/topologies/#centralized[deployment topology] and https://vector.dev/docs/reference/configuration/sinks/[sinks].
```

modules/concepts/pages/logging.adoc
```diff
@@ -1,32 +1,57 @@
...
+You configure your logging settings in the resource describing the cluster (ZookeeperCluster in this example), seen in the top left in the diagram (see the <<configuration, configuration>> section below).
+The operator reads this resource and creates the appropriate log configuration files in the ConfigMap which also holds other product configuration.
+The ConfigMap is then mounted into the containers.
+The ZooKeeper Pod has three containers: The `prepare` sidecar container, the `zookeeper` container and the `vector` https://vector.dev/docs/setup/deployment/roles/#sidecar[sidecar container].
+All logs get written into a shared mounted directory, from which the Vector agent reads them and sends them to the vector https://vector.dev/docs/setup/deployment/roles/#aggregator[aggregator].
+
+=== Aggregator and sinks
+
+The aggregator is configured to use one or multiple https://vector.dev/docs/reference/configuration/sinks/[sinks] (for example OpenSearch, Elasticsearch, WebHDFS), it sends all logs to all sinks.
```
siegfriedweber, line 43: Example configurations for OpenSearch/Elasticsearch and WebHDFS would be helpful in this documentation. The link to the logging demo is not enough because there is also no configuration shown.

fhennig: I agree that this would be useful, but I think it is outside of the scope of #470. I have made a ticket for it though: https://github.com/stackabletech/documentation/issues/476

siegfriedweber (dismissed):

fhennig:
Actually, maybe we should also fix this: https://github.com/stackabletech/zookeeper-operator/issues/736

the docs now claim that config is always the same, and I'm using zookeeper explicitly as an example ...

fhennig:
it's fixed, but now it doesn't work with nightly because of a vector restructuring, gotta look into it.

modules/tutorials/examples/logging-aggregator/main.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+echo "Installing ZooKeeper Operator"
+# tag::zk-op[]
+stackablectl release install -i secret -i commons -i zookeeper 23.7
+# end::zk-op[]
+
+# tag::vector-agg[]
+helm install -f vector-aggregator-values.yaml vector-aggregator vector/vector
```
siegfriedweber, line 9: This script fails because the Vector aggregator is not ready when the logs are queried.

```suggestion
helm install --wait -f vector-aggregator-values.yaml vector-aggregator vector/vector
```

modules/tutorials/examples/logging-aggregator/main.sh
```diff
@@ -0,0 +1,32 @@
...
+# tag::vector-discovery[]
+kubectl apply -f vector-aggregator-discovery.yaml
+# end::vector-discovery[]
+
+# tag::zk[]
+kubectl apply -f zookeeper.yaml
+# end::zk[]
+
+# TODO make this somehow a testable thing
+# maybe get the logs from the agg. and grep for a specific line?
```
siegfriedweber, line 21: If you wait until the ZooKeeper instances are ready then this script is a working test.

```suggestion
kubectl rollout status statefulset simple-zk-server-default --timeout=5m
kubectl wait \
    --for=jsonpath='.status.readyReplicas'=3 \
    --timeout=5m \
    statefulsets.apps/simple-zk-server-default
```

siegfriedweber (changes requested):

modules/concepts/images/logging_architecture.drawio.svg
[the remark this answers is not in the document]
fhennig: https://github.com/stackabletech/documentation/pull/471/commits/402f843d89dae3e857bd0763d9e8461057138594

modules/tutorials/examples/logging-aggregator/main.sh
fhennig: @siegfriedweber Why the "rollout status" and the "kubectl wait", shouldn't one of them be enough?

siegfriedweber: `kubectl rollout status …` waits until the StatefulSet is deployed and the pods are running and `kubectl wait …` waits until   the pods are ready. If `kubectl rollout status …` is left out then `kubectl wait …` fails immediately because it cannot find the given StatefulSet. If `kubectl wait …` is left out then the pods are running but not necessarily ready.

modules/tutorials/examples/logging-aggregator/main.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+echo "Installing ZooKeeper Operator"
+# tag::zk-op[]
+#stackablectl release install -i secret -i commons -i zookeeper 23.7
```
siegfriedweber, line 5: I would remove the comment. It is shown on the documentation web page and it is confusing to have it there.

modules/tutorials/examples/logging-aggregator/main.sh
```diff
@@ -0,0 +1,37 @@
...
+stackablectl op in secret commons zookeeper
```
siegfriedweber, line 6: I would always use the descriptive variants of the command parameters in the documentation:

```suggestion
stackablectl operator install secret commons zookeeper
```

This does not exceed the line length.

modules/tutorials/examples/logging-aggregator/main.sh
```diff
@@ -0,0 +1,37 @@
...
+# end::zk-op[]
+
+# tag::vector-agg[]
+helm install --wait -f vector-aggregator-values.yaml vector-aggregator vector/vector
```
siegfriedweber, line 10: On large screens, only 75 characters are shown per line. To avoid that the user has to scroll horizontally in a code block, this limit should not be exceeded.

Perhaps it makes sense to indent the lines only by two spaces to save even more space (especially on mobile devices).

I also replaced `-f` with `--values`. It takes more space but long names are often more readable.

```suggestion
helm install \
  --wait \
  --values vector-aggregator-values.yaml \
  vector-aggregator vector/vector
```

modules/tutorials/pages/logging-vector-aggregator.adoc
```diff
@@ -0,0 +1,113 @@
...
+== Watch the logs
+
+During startup, ZooKeeper already prints out log messages.
+Vector was configured to print the aggregated logs to `stderr`, so if you look at the logs of the Vector pod, you will see the ZooKeeper logs:
+
+
+[source,bash]
+include::example$logging-aggregator/main.sh[tag=grep]
+
+You should see a JSON oject per ZooKeeper replica printed that look like
```
siegfriedweber, line 81:
```suggestion
You should see a JSON object per ZooKeeper replica printed that looks like
```

modules/tutorials/pages/logging-vector-aggregator.adoc
```diff
@@ -0,0 +1,113 @@
...
+  "source_type": "vector",
+  "timestamp": "2023-11-06T10:30:40.223Z"
+}
+
+The JSON object contains a timestamp, the log message, log level and some additional information.
+
+You can see the same log line in the log output of the ZooKeeper container:
+
+[source]
+kubectl logs --container=zookeeper simple-zk-server-default-0 | grep "zookeeper.version="
```
siegfriedweber, line 104:
```suggestion
kubectl logs \
  --container=zookeeper simple-zk-server-default-0 \
  | grep "zookeeper.version="
```

siegfriedweber (changes requested):

modules/tutorials/examples/logging-aggregator/main.sh
fhennig: This is the "real" command we should use, but with 23.11. Otherwise we get nightly operators. But we can only do this once 23.11 is released.

modules/tutorials/examples/logging-aggregator/main.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# tag::vector-discovery[]
+kubectl apply -f vector-aggregator-discovery.yaml
+# end::vector-discovery[]
+
+# tag::zk[]
+kubectl apply -f zookeeper.yaml
+# end::zk[]
+
+sleep 10
```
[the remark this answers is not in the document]
fhennig: this is to give the API server time to actually create the objects from zookeeper.yaml. Without a brief sleep, the "rollout status" fails because the statefulset isn't even there yet.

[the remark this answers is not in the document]
siegfriedweber: Instead of sleeping, you could use a server-side apply:

```shell
kubectl apply -f zookeeper.yaml --server-side
```

modules/tutorials/examples/logging-aggregator/main.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+echo "Installing ZooKeeper Operator"
+# tag::zk-op[]
+#stackablectl release install -i secret -i commons -i zookeeper 23.7
```
siegfriedweber, line 5: Please install the listener-operator too, because as one of the fundamental operators, it should be part of every installation.

modules/tutorials/examples/logging-aggregator/main.sh
fhennig: [956e728](https://github.com/stackabletech/documentation/pull/471/commits/956e728f8be31b5c8255ac324ff35976b9b68ad5)

modules/tutorials/examples/logging-aggregator/main.sh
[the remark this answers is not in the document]
fhennig: https://github.com/stackabletech/documentation/pull/471/commits/c1a8bbe33285d3389f83ce0b687c7203de410961

modules/tutorials/examples/logging-aggregator/main.sh
fhennig: https://github.com/stackabletech/documentation/pull/471/commits/87628dfce414a1344330776449b92eca7e428060

siegfriedweber (approved):
LGTM

PR #474, opened 2023-10-19
New menu structure & lifecycle policies

### LINKED ISSUE #475: Document our policies around upgrade paths
We decided that - for now - we will only support upgrading from one version to the next. Skipping an intermediary release is not currently supported.

This policy needs to be documented. It doesn't have to be much.

Can be something like this

```
= Upgrade Policy

Skipping SDP releases while upgrading is not currently supported.
The only upgrade path we support is from one version to the next.
```

fhennig:
fixes #475


modules/ROOT/pages/policies.adoc
```diff
@@ -47,26 +45,26 @@ image:full_support_scenario_2.png[Full Support Scenario 2]
...
+We may release patches for other reasons.
 
 image:maintenance_phase.png[Maintenance Phase]
 
-=== General
-
-We may release patches for other reasons.
+== Upgrade policy
 
 Customers are expected to upgrade their SDP environment to the most current supported patch/micro (`23.4.z`) version.
```
Techassi, line 59: I would just call it patch version. The SemVer spec also uses the term `patch`.

modules/ROOT/pages/policies.adoc
```diff
@@ -47,26 +45,26 @@ image:full_support_scenario_2.png[Full Support Scenario 2]
...
-Make sure to follow our release notes to learn of valid upgrade paths, some intermediate updates might be necessary.
+Skipping SDP releases while upgrading is not currently supported.
```
Techassi, line 60:
```suggestion
Skipping SDP releases while upgrading is currently not supported.
```

supplemental-ui/partials/navbar.hbs
```diff
@@ -30,4 +29,20 @@
...
 </div>
+<div class="navbar-sub-item drop-down">
+    Management
+    <div class="drop-down-content">
+        <a class="drop-down-item" href="{{{ relativize (versioned "management" page "cockpit/index.html") }}}">Cockpit</a>
+        <a class="drop-down-item" href="{{{ relativize (versioned "management" page "stackablectl/index.html") }}}">stackablectl</a>
+    </div>
+</div>
+<div class="navbar-sub-item drop-down">
+    Product Information
```
Techassi, line 40: Maybe a better term would be `Policies and Licenses`. I would also suggest to move `Releases Notes` to the platform menu.

Techassi (changes requested):
Some minor nits.

modules/ROOT/pages/policies.adoc
fhennig: Earlier in the document it is mentioned that CalVer calls this part of the version the "micro" bit: https://calver.org/

That's why patch and micro are mentioned

supplemental-ui/partials/navbar.hbs
fhennig: Thanks for the suggestion, I like it and did it.

modules/ROOT/pages/policies.adoc
Techassi: Ah alright, then it's fine I guess.

modules/ROOT/pages/policies.adoc
```diff
@@ -20,21 +19,20 @@ All these policies may evolve as we learn what works best for our customers and
...
-== Stackable Data Platform Lifecycle Policy
+== Stackable Data Platform lifecycle policy
 
 NOTE: This policy concerns releases of our platform as a whole and how long and to which extent we support each version.
 
 We do releases of our Stackable Data Platform.
 These releases get a name based on the year and month they have been released in (e.g. `23.4`, `23.7`, also called https://calver.org/[CalVer]). This name does not follow Semantic Versioning (https://semver.org/[SemVer]). We may release patches for a release, which then follow the PATCH naming semantics of SemVer (e.g. `23.4.1`) or the _Micro_ name from CalVer. See below for our policy on patches for the SDP.
 
-We do support an SDP release for a specific time after its release.
+We support an SDP release for a specific time after its release.
```
Techassi, line 29:
```suggestion
We support a SDP release for a specific amount of time after its initial release.
```

I'm not quite sure if `a` or `an` is correct here. Technically `a` is correct, but sounds a little funny in context.

Techassi (changes requested):
Final nitpick.

modules/ROOT/pages/policies.adoc
fhennig: it's about the _sound_ so it'd be "an" here because it's "an _es dee pee_ release". But I like the other stuff!

Techassi: Yeah, I know. That's why I was a little unsure.

Techassi (approved):

PR #479, opened 2023-11-01
Glossary

razvan:


modules/concepts/pages/glossary.adoc
```diff
@@ -0,0 +1,8 @@
+= Glossary
+
+== Stacklet
+
+A custom resource for a product and all objects that are owned by it.
```
nightkr, line 5: This explanation feels very Kubernetes-centric, rather than user/goal-centric.

I'd start with something like "A Stacklet is an installation of a product.". That we implement that by creating various Kubernetes objects is a later detail.

nightkr: That said, good initiative on getting *something* up here!

fhennig:
more things to cover:

- role
- role group
- service discovery configmap
- authenticationclass
- secretclass
- listener etc.
- reconciliation
- opa
- vector

fhennig:
I know the syntax in the document looks horrible now but I think the page looks great!

fhennig (dismissed):

modules/concepts/pages/glossary.adoc
razvan: I had forgotten about this. I agree. Felix made some changes.

fhennig (approved):

PR #486, opened 2023-11-20
docs: Add listener-operator where missing

siegfriedweber:
The listener-operator is now one of the fundamental operators, along with the commons- and secret-operators, and should be mentioned accordingly everywhere.


maltesander (approved):
LGTM!

PR #489, opened 2023-11-27
Add redirect from versioned link to stable

fhennig:
Currently there is no way to link to the latest version by its actual version number, which is annoying. There is only the 'stable' link (i.e. https://docs.stackable.tech/home/stable/concepts/product_image_selection). Previous versions are linkable by their version number (i.e. https://docs.stackable.tech/home/23.4/concepts/product_image_selection). 

This PR adds a redirect from the actual version to 'stable'.

**Why redirect from the number to stable, and not the other way round?** Where do URLs actually end up? Users might send each other links, they are just used in the moment and then the URL doesn't really matter. For bookmarking a URL, sometimes bookmarking a link that always points to the latest docs might be better, sometimes you might want a stable link. Both are possible, just the default would be to link 'stable'. **The main reason is SEO** - Google will index the 'stable' links, and when we release a new version, the pages will update, instead of google having to find the new version number. If a new version is released, google will find the now latest-1 version on a new indexing run and will index it too, but it will also update the 'stable' pages.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -7,13 +7,19 @@ site:
   keys:
     enable_tracking: true
 
+# URL config settings.
+# docs: https://docs.antora.org/antora/latest/playbook/configure-urls/
 urls:
-  # The symbolic name of the latest stable version. This is used to generate redirects to the latest version.
+  # The symbolic name of the latest stable version, and how to redirect from/to it.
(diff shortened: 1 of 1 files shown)


vsupalov (approved):
Redirect results in 302s (moved temporarily), allows for stable external links to versions and otherwise should not interfere with existing content discoverability 👍 LGTM

PR #482, opened 2023-11-14
Release notes for SDP release 23.7

adwk67:

PR #490, opened 2023-11-28
Aadapt ldap tutorial to 23.11 stack changes

maltesander:


modules/tutorials/examples/ldap-auth/trino-opa-bundle-snippet.yaml
```diff
@@ -5,9 +5,14 @@ data:
...
+      some i
+      input.action.filterResources[i]
+      is_alice
     }
 
-    allow {
-      input.context.identity.user == "bob"
-    }
\ No newline at end of file
+    is_alice() {
```
adwk67, line 16: Not sure if this is more correct than the other open PR. See here: https://github.com/stackabletech/documentation/pull/482/files#diff-e8fbe72e226a4c1a31030b2c83a56e67abb2bdc616fd9efa263eb844ce562ce7
You can judge that better than me!

adwk67 (approved):
Not sure about the clash with opa-snippet but you're the best judge of what is better. LGTM.

modules/tutorials/examples/ldap-auth/trino-opa-bundle-snippet.yaml
maltesander: We just keep alice to showcase that bob cannot do anything then.

PR #497, opened 2023-12-04
post-release notes: known issues and OS deployment

adwk67:


modules/ROOT/pages/release-notes.adoc
```diff
@@ -6,7 +6,7 @@ including all components of the platform at a specific version.
 
 == Release 23.11
 
-This release introduces two further elements of platform stability - pod disruption budgets and graceful shutdown specifications - as well as a new management tool and updated product version support.
+This release introduces two further elements of platform stability - pod disruption budgets and graceful shutdown specifications - as well as a new management tool and updated product version support. Any known issues related to this release are listed below in a separate section.
```
NickLarsenNZ, line 9: I think there should be a link to go straight there (even though it is on the right-hand side navigation).

I'm unsure how to link to the right one though, as the anchor is `_known_issues` which will collide with future _Known Issues_.

NickLarsenNZ (approved):
Left a comment, but approving anyway

PR #509, opened 2023-12-14
Link to CRD docs

fhennig:
Add links to the CRD docs in various places. Delete old hand-written reference pages.


supplemental-ui/partials/navbar.hbs
```diff
@@ -6,6 +6,7 @@
         <a class="drop-down-item" href="{{{ relativize (versioned "home" page "concepts/index.html") }}}">Concepts</a>
         <a class="drop-down-item" href="{{{ relativize (versioned "home" page "demos/index.html") }}}">Demos</a>
         <a class="drop-down-item" href="{{{ relativize (versioned "home" page "tutorials/index.html") }}}">Tutorials</a>
+        <a class="drop-down-item" href="{{{ relativize (versioned "home" page "reference/index.html") }}}">CRD Reference</a>
```
sbernauer, line 9: Can we directly point to https://crds.stackable.tech/nightly here? Saves a click :)
We could also get rid of `modules/reference/pages/index.adoc` in the nav.adoc entirely and only leave it in to let the page-aliases work

sbernauer: > Maybe, if it's not super urgent, we can postpone this as an improvement for the future, and just move the bits we have now?

Sure thing! I've heard you had fiddled around after this comment, also happy to review the changes :)

sbernauer (approved):

sbernauer (approved):
Awesome, thanks!

sbernauer (approved):

PR #518, opened 2024-01-04
docs: Add code style guide

Techassi:


modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
+= Source code style guide
+
+:k8s-quantity: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/
+
+== Long versus abbreviated variable names
+
+We use long (no abbreviated) variables names for function parameters and locally declared variables.
+Short (even single letter) variable names are allowed in lambdas (closures) and one-liners.
+The usage of well-known acronyms like CPU, TLS or OIDC are allowed.
```
NickLarsenNZ, line 9: Should this apply to all identifiers or at least variables and functions?

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+Examples of *incorrect* code for this rule:
+
+[source,rust]
+----
+let param = Some("foo");
+let buf = &[];
+
+fn func(elems: Vec<String>) {}
```
NickLarsenNZ, line 18: the function name in this example breaks the rule (I know this is about variable names, but it still draws attention to the inconsistency)

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+:k8s-quantity: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/
+
+== Long versus abbreviated variable names
+
+We use long (no abbreviated) variables names for function parameters and locally declared variables.
+Short (even single letter) variable names are allowed in lambdas (closures) and one-liners.
+The usage of well-known acronyms like CPU, TLS or OIDC are allowed.
+
+Examples of *incorrect* code for this rule:
```
NickLarsenNZ, line 11: I think it could be useful to have a visual indicator for examples of correct vs incorrect? Eg:

> :x: Examples of *incorrect* code for this rule:

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+let oidc_provider = OidcProvider {};
+let tls_settings = TlsSettings {};
+----
+
+== Captured versus uncaptured format string identifiers
+
+We generally use captured format string identifiers.
+The identifiers should be concise but not abbreviated.
+Mix-and-matching of captured and uncaptured format string identifiers must be avoided.
```
NickLarsenNZ, line 44: ... "To remove ambiguity, and to decouple argument order from the text (which can lead to incorrect text when the wording is changed and `{}` are reordered while the arguments aren't)."

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+We place no restriction on named format string identifiers.
+Both options below are considered valid.
+
+[source,rust]
+----
+let greetee = "world";
+
+format!("Hello, {greetee}!");
+format!("Hello, {greetee}!", greetee = "world");
```
NickLarsenNZ, line 72: Maybe `world` can be changed to something else, like `universe` to really make the distinction obvious?

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+----
+
+== Newlines between struct fields and enum variants
+
+We add newlines to struct fields and enum variants when they include additional information like documentation comments or attributes.
```
NickLarsenNZ, line 77: ... "because the variants can become difficult to read" (or something something readability)

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
+= Source code style guide
+
+:k8s-quantity: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/
+
+== Long versus abbreviated variable names
+
+We use long (no abbreviated) variables names for function parameters and locally declared variables.
```
NickLarsenNZ, line 7: ... "to avoid ambiguity"

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+// Well-known acronyms
+
+let oidc_provider = OidcProvider {};
+let tls_settings = TlsSettings {};
+----
+
+== Captured versus uncaptured format string identifiers
+
+We generally use captured format string identifiers.
```
NickLarsenNZ, line 42: Is there a case we wouldn't? Or should we make it a hard rule going forward)?

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+  /// Indicates that we failed to bar.
+  #[snafu(display("failed to bar"))]
+  Bar,
+  Baz,
+}
+----
+
+== Error handling
+
+We use `snafu` for all error handling in library *and* application code.
```
NickLarsenNZ, line 127: I think we need to give a reason for this as opposed to custom error objects, or using `thiserror` which is fairly ubiquitous.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+    user_name: user.name,
+  });
+}
+----
+
+== Error messages
+
+All our error messages must start with a lowercase letter and must not end with a dot.
+Additionally, critical (unrecoverable) errors should start with "failed ...".
+Potentially recoverable errors should start with "unable ...".
```
NickLarsenNZ, line 183: nit: I think we should say "failed to ..." and "unable to ...". IMO, the "to" leaves less guess-work for the reader.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+Examples of *incorrect* code for this rule:
+
+[source,rust]
+----
+#[derive(Snafu)]
+enum Error {
```
NickLarsenNZ, line 190: _marker comment about singular named enums (not plural). Reason: only one variant is valid `Color::Red` and not `Colors::Red`_

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+[source,rust]
+----
+let memory: MemoryQuantity = "100Mi".parse();
+let memory: MemoryQuantity = "1Gi".parse();
+let memory: MemoryQuantity = "1536Mi".parse();
+let memory: MemoryQuantity = "10Gi".parse();
+----
+
+=== Resources measured in CPU fractions
```
NickLarsenNZ, line 268: Missing "Examples of incorrect code for this rule:"

NickLarsenNZ (changes requested):
Really nice, thanks for this! I was just recently looking for the rule about variants which I knew I'd seen somewhere (but was hidden in a private document).

I think adding reasons for each rule would be good,
- it helps the reader understand why the rule exists
- over time, reasons can become invalid and we can replace/remove the rule instead of it becoming dogma

modules/contributor/pages/code-style-guide.adoc
Techassi: You are right. It seems a good idea to extend that section to include function names and also constants. Is there anything other which might be worth including?

modules/contributor/pages/code-style-guide.adoc
Techassi: I will rename the function.

modules/contributor/pages/code-style-guide.adoc
Techassi: Good idea. I thought about using admonitions, but the AsciiDoc ones don't support custom titles. Using emojis seems like a good alternative.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
+= Source code style guide
+
+:k8s-quantity: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/
+
+== Long versus abbreviated variable names
+
+We use long (no abbreviated) variables names for function parameters and locally declared variables.
+Short (even single letter) variable names are allowed in lambdas (closures) and one-liners.
```
nightkr, line 8: I'd say this is more about whether the meaning of the variable can be easily parsed from the surrounding context.

modules/contributor/pages/code-style-guide.adoc
nightkr: "We use unabbreviated identifier names."? No need to enumerate every single case where it should apply (especially since we handle exceptions separately anyway).

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+----
+
+Examples of *correct* code for this rule:
+
+[source,rust]
+----
+let parameter = Some("foo");
+let buffer = &[];
+
+fn func(elements: Vec<String>) {}
```
nightkr, line 28: Honestly... I have to disagree and say that I find the "wrong" far more pleasant to read. I guess to my brain `buf`, `param`, and `elem` are all in that bucket of so-well-known-it-gets-kinda-silly.

modules/contributor/pages/code-style-guide.adoc
nightkr: You can't use it for substitutions that aren't straightforward identifiers (function calls, paths like `foo.bar` or `Foo::BAR`, etc).

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+// Well-known acronyms
+
+let oidc_provider = OidcProvider {};
+let tls_settings = TlsSettings {};
+----
+
+== Captured versus uncaptured format string identifiers
+
+We generally use captured format string identifiers.
+The identifiers should be concise but not abbreviated.
```
nightkr, line 43: Disagree somewhat. rustfmt breaks when string literals create overly long lines, so avoiding that takes precedence imo.

modules/contributor/pages/code-style-guide.adoc
nightkr: Oh... I think I see the confusion here.

I'd say "named format string substitutions", "captured" implies variables captured from the environment (that is, `let foo = "bar"; format!("{foo}")` rather than `format!("{foo}", foo = "bar")`.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+Mix-and-matching of captured and uncaptured format string identifiers must be avoided.
+
+Examples of *incorrect* code for this rule:
+
+[source,rust]
+----
+format!("Hello, {}!", "world");
```
nightkr, line 50: Disagreed, this is fine when there is a single trivial substitution. `format!("{}", foo())` is fine, `format!("This is the {}{} time you have done this {}!", 4, "th", "this week")` should definitely use named substitutions instead.

modules/contributor/pages/code-style-guide.adoc
nightkr, line 77: I wonder if this is an editor settings thing? If anything I find doccomments or attributes to be a separator when reading it.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+enum Error {
+  /// Indicates that we failed to foo.
+  #[snafu(display("failed to foo"))]
+  Foo,
+
+  /// Indicates that we failed to bar.
+  #[snafu(display("failed to bar"))]
+  Bar,
+  Baz,
```
nightkr, line 121: This example really highlights (to me) how hilariously backwards this rule is.

modules/contributor/pages/code-style-guide.adoc
nightkr: I also don't like at all how this further increases the "noise cost" of adding doccomments.

modules/contributor/pages/code-style-guide.adoc
nightkr, line 183: Not sure about this. First, I'd say there are two kinds of errors, "we tried to do X and failed", and "we failed because of X". "X happened" is fine to me, _provided_ that it's wrapped by "we tried to do X"-style errors.

Second, I don't think "recoverable" vs "unrecoverable" makes sense. Most errors are recoverable in some sense. "Transient" vs "requires intervention" might be _closer_, but it's also often hard for us to know without the wider context. We might be unable to reach a server because it's restarting, because its disk suffered the click of death, or because we have the wrong URL. All of those require vastly different responses from different people, but they look identical on our end.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+  Foo,
+
+  #[snafu(display("unable to bar"))]
+  Bar,
+}
+----
+
+== Naming optional function parameters and variables
+
+Optional function parameters and variables containing `Option` must not use any prefixes or suffixes to indicate the value is of type `Option`.
```
nightkr, line 218: Agree with this _most_ of the time, but we also have a few cases where we want to explicitly handle "not found" cases (`Client::get_opt`).

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+
+=== Resources measured in bytes
+
+Examples of *incorrect* code for this rule:
+
+[source,rust]
+----
+// Biggest matching unit
+let memory: MemoryQuantity = "1.5Gi".parse();
+let memory: MemoryQuantity = "1Gi".parse();
```
nightkr, line 248: `1Gi` is listed as both correct and incorrect.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+let memory: MemoryQuantity = "10Gi".parse();
+----
+
+=== Resources measured in CPU fractions
+
+[source,rust]
+----
+// Biggest matching unit
+let memory: CpuQuantity = "500m".parse();
+let memory: CpuQuantity = "1".parse();
```
nightkr, line 274: Same here for `1`.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,293 @@
...
+Examples of *correct* code for this rule:
+
+[source,rust]
+----
+let tls_settings: Option<TlsSettings> = None;
+----
+
+== Specifying resources measured in bytes and CPU fractions
+
+We follow the Kubernetes convention described {k8s-quantity}[here].
```
nightkr, line 238: How about a style guide entry against using substitutions for link URLs used in one place?

nightkr:
Also, this conventional commits obsession is getting inane. This is the docs repository. It contains nothing but docs. Of course every single PR is going to be about docs!

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: Yeah, I was just assuming emojis

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: I do to tbh, but I'd say that comes from reading enough C where that's the norm. I still think the rule is valid. Is a change to the example needed? (IMO, no, I think it conveys the point well enough).

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: I'm not sure I understand what you mean. The `_opt` seems unnecessary. Could you elaborate?

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: I like that (what you suggest), unless they are super long and ruin the readability in plain text form.

nightkr: Honestly I'd still rather like to be able to just follow the link in the plain-text variant too.

NickLarsenNZ:
> Also, this conventional commits obsession is getting insane. This is the docs repository. It contains nothing but docs. Of course every single PR is going to be about docs!

@nightkr there's also ci (rareley changes), but yeah, the limited variance of types of changes makes conventional commits a bit too much. But also, it becomes habit too, so can we allow them but not enforce them here?

modules/contributor/pages/code-style-guide.adoc
nightkr: We use `::get()` for the common case where we expect the object to exist, and "not found" is a hard error that we should fail on. `::get_opt()` is a special-case for when we want to handle missing objects gracefully.

NickLarsenNZ: Ok. Not sure I have come across that yet. I would expect `get()` to return an option, or in the case of it being fallable, a `Result<Option<T>>`, rather that two separate methods.

But I'm also thinking more like _get from a DB_ where you wouldn't expect something to exist at code/compile time.

modules/contributor/pages/code-style-guide.adoc
Techassi: The `buf` versus `buffer` example was directly taken from the code style table in our arch meeting sheet.

modules/contributor/pages/code-style-guide.adoc
Techassi: Yeah, there is some overlap. I only included parts of the styles discussed in the arch meeting sheet. I will add the others to reduce the confusion.

modules/contributor/pages/code-style-guide.adoc
Techassi: Changed in 5fa9326. It also experiments with a different view for correct/incorrect usages mentioned here https://github.com/stackabletech/documentation/pull/518#discussion_r1442645046

modules/contributor/pages/code-style-guide.adoc
Techassi: Renamed in 5fa9326

modules/contributor/pages/code-style-guide.adoc
Techassi: I figured out a way to use admonitions. Looks really nice in my opinion. It requires some CSS adjustments which I will add in a separate PR in https://github.com/stackabletech/documentation-ui

modules/contributor/pages/code-style-guide.adoc
Techassi: Nice idea!

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: I like this!

modules/contributor/pages/code-style-guide.adoc
Techassi: Changed in a979df4

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: Even with syntax highlighting, I find them too noisy without the line break.
I can't see a rustfmt option for it, but I feel like I have had _something_ automatically break them up for me in vscode.

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: I'm not sure I follow. What's the problem in this example?

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: Fair

modules/contributor/pages/code-style-guide.adoc
Techassi: Added in e0ece4f

modules/contributor/pages/code-style-guide.adoc
Techassi: Added in 08d9383

modules/contributor/pages/code-style-guide.adoc
Techassi: Added in 245f1e4

modules/contributor/pages/code-style-guide.adoc
Techassi: Fixed in bbb84e9

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,419 @@
...
+----
+#[derive(thiserror::Error)]
+enum Error {
+  #[error("failed to read config file")]
+  FileRead(#[from] std::io::Error)
+}
+
+fn config_file(user: User) -> Result<(), Error> {
+  std::fs::read_to_string(user.file_path)?;
+}
```
maltesander, line 223: Why is there a thiserror example (and then not in the correct example)?

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -0,0 +1,419 @@
...
+
+We place no restriction on named format string identifiers.
+Both options below are considered valid.
+
+[source,rust]
+----
+let greetee = "world";
+
+format!("Hello, {greetee}!");
+format!("Hello, {greetee}!", greetee = "universe");
```
maltesander, line 345: Imho i would be strict about going for the second one if the var occurs multiple times?

modules/contributor/pages/code-style-guide.adoc
maltesander: Stupid question: Why is that incorrect :D ?

modules/contributor/pages/code-style-guide.adoc
Techassi: This just shows that using thiserror is considered incorrect.

modules/contributor/pages/code-style-guide.adoc
sbernauer: Honestly I'm with @nightkr here

sbernauer (commented):
Thanks a lot, LGTM % comments

modules/contributor/pages/code-style-guide.adoc
Techassi: Changed in 430f2cc

modules/contributor/pages/code-style-guide.adoc
Techassi: Yeah, I will try to re-word it. Simple substitution should be allowed, but complex one should use named identifiers.

modules/contributor/pages/code-style-guide.adoc
Techassi: I think there are some rules in place, but we cannot use them. I researched this a few weeks back.

modules/contributor/pages/code-style-guide.adoc
Techassi: Yeah, it doesn't make sense in the current context. I will adjust this.

modules/contributor/pages/code-style-guide.adoc
Techassi: *Gonna use this link twice so that Natalie doesn't complain :)* Just kidding, I can inline the link. I just find these super long ones ugly while looking at the plain text file.

modules/contributor/pages/code-style-guide.adoc
Techassi: Changed in 499cbb1

modules/contributor/pages/code-style-guide.adoc
Techassi: Changed in 499cbb1

modules/contributor/pages/code-style-guide.adoc
Techassi: Changed in f256324

modules/contributor/pages/code-style-guide.adoc
Techassi: Changed in f256324

modules/contributor/pages/code-style-guide.adoc
Techassi: Mentioned in 1f45821

modules/contributor/pages/code-style-guide.adoc
Techassi: > if the var occurs multiple times

What do you exactly mean with that statement? Multiple times in the same `format!()` call or multiple times in the current scope? Or something else entirely?

modules/contributor/pages/code-style-guide.adoc
Techassi: > Not sure about this. First, I'd say there are two kinds of errors, "we tried to do X and failed", and "we failed because of X". "X happened" is fine to me, provided that it's wrapped by "we tried to do X"-style errors.

I agree. This mostly should be covered by a "failed to ..." error, followed by an "unable to ..." error. In a more natural way of saying things: "Failed to X and thus we were unable to Y".

> Second, I don't think "recoverable" vs "unrecoverable" makes sense. Most errors are recoverable in some sense. 

Yes, I thought about that as well when writing this paragraph. I agree that these terms are not the right ones. How would you call these two types of errors?

> "Transient" vs "requires intervention" might be closer, but it's also often hard for us to know without the wider context.

Do you have any suggestions on the wording in our rules so that we more *easily* decide which variant we want to use in code?

modules/contributor/pages/code-style-guide.adoc
Techassi: I can include that in the rule. At this point it might even be clever to split the formatting into multiple parts.

modules/contributor/pages/code-style-guide.adoc
maltesander: The first one like:
```
format!("Hello {name}, hello again {name}", name = greetee);
```
but no strong opinion... could clear up the variable context but that should do the variable name itself if you adhere to this style guide ;)

Techassi: Added in 1245908

modules/contributor/pages/code-style-guide.adoc
Techassi: Added in 1924e9b

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: I agree.

There's a rule of thumb I tend to stick to - just realised it's an Uncle Bobism:

> The shorter the scope the shorter the variable names, ...
> Source: [X, formally Twitter](https://twitter.com/unclebobmartin/status/360029878126514177?lang=en)

<details><summary>example</summary>

So like `i` as an iterator index for a short scoped for loop, or `volume_name`in the context of some kind of volume builder, or `tls_volume_name` when outside of that scope.

</details>

Techassi: I really like the example. I will try to incorporate it into our rule.

Techassi: Added in aee5be0

modules/contributor/pages/code-style-guide.adoc
nightkr: `Foo` and `Bar` are already clearly separated by the comments, while `Baz` blends into `Bar`.

IMO a whole "block" should be either commented or not, so this would be okay:

```rust
enum X {
  /// foo
  Foo,

  /// bar
  Bar,

  A,
  B,
  C,
}
```

That said, in most cases I'd also consider single uncommented variants in an otherwise-commented enum to be a smell.

modules/contributor/pages/code-style-guide.adoc
nightkr: > Do you have any suggestions on the wording in our rules so that we more easily decide which variant we want to use in code?

That's my point, I'm not sure we *can* always judge that for sure. It depends on context that we don't have.

modules/contributor/pages/code-style-guide.adoc
Techassi: > while Baz blends into Bar.

Yes, this currently allowed. I'm also happy to make this invalid.

> That said, in most cases I'd also consider single uncommented variants in an otherwise-commented enum to be a smell.

Agreed. Ideally *all* variants and fields should be documented. I can add this as a note to the rule. Same goes for the point above.

modules/contributor/pages/code-style-guide.adoc
Techassi: Should we then only say that we use error messages which either start with "failed to..." and "unable to ..." without making any assumptions on when to use them?

modules/contributor/pages/code-style-guide.adoc
nightkr: > Agreed. Ideally all variants and fields should be documented.

Eh, kinda... There are still utterly useless comments that I'd rather do without, like:

```rust
/// do baz
fn baz()
```

The comment, if it exists, should add _something_ that the name and types alone can't convey. Like, don't do this:

```rust
/// x must be UTF-8
fn baz(x: &[u8])
```

when you could do this:

```rust
fn baz(x: &str)
```

Techassi: Agreed, again. I will add a statement to make sure we only add good quality comments to the fields / variants (if they are required at all).

modules/contributor/pages/code-style-guide.adoc
nightkr: I don't know. I'd rather focus on coming up with rules of thumbs for when to follow which style. I don't think "YOU MUST FOLLOW ONE OF THESE STYLES (although it's so watered down that neither actually means anything)" is helpful for anyone.

Techassi: Alright, I will come up with these rules of thumb!

modules/contributor/pages/code-style-guide.adoc
NickLarsenNZ: For enum variants that don't have a comment (because they are perhaps too obvious), they should still be separated if they have a macro above them too.

Eg: All of these being valid:

```rust
enum Color {
  Red,
  Green,
  Blue,
}

#[derive(Debug, Snafu)]
enum Error {
  #[snafu(display("failed to stat file"))]
  StatFile,

  #[snafu(display("failed to write to file"))]
  WriteFile,
}

enum Operation {
  /// A command that takes no arguments
  Command,

  /// A command which takes the specified arguments
  CommandWithArgs(Args),
}
```

Techassi: Addressed in ae09e54

modules/contributor/pages/code-style-guide.adoc
Techassi: Added in fae529a

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -204,6 +220,11 @@ enum Error {
 
 ====
 
+Any single uncommented variants or fields in an otherwise-commented enum or struct is considered to be a smell.
+If any of the items is commented, all items should be.
+It should however also be noted that there is no requirement to comment fields or variants.
+Comments should only be added if they provide additional information not available from context.
```
NickLarsenNZ, line 226: praise: :smile:

NickLarsenNZ (approved):
LGTM

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -204,6 +220,11 @@ enum Error {
 
 ====
 
+Any single uncommented variants or fields in an otherwise-commented enum or struct is considered to be a smell.
+If any of the items is commented, all items should be.
```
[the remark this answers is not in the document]
Techassi: I created #521 to tackle this.

PR #522, opened 2024-01-09
ADR: CA bundle injection for webhooks

maltesander:
Refinement 05.02.2024

* On OpenShift, OLM deployments manages CAs it's self
* We don't consider it acceptable to require customers to manage a Cert Manager deployment just for the purpose of registering webhooks.

- [ ] We need to have a feature flag that can be turned off on platforms that manage CAs themselves.
- [ ] Make the key reader configurable because there may be different key formats. The OLM spike has shown that OLM generates keys in `ec` format.

preview link: https://deploy-preview-522--stackable-docs.netlify.app/home/nightly/contributor/adr/adr033-foundation-webhooks-ca-bundle


modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,211 @@
...
+Technical Story: https://github.com/stackabletech/issues/issues/361
+
+== Context
+
+There are many use cases for the future development of the SDP that involve validating, mutating and conversion webhooks. Tasks that must be tackled in the near future include:
+
+- The proper versioning of our CRDs and the possibility of smoothly up- and downgrading between versions including breaking changes (see https://github.com/stackabletech/documentation/issues/273)
+- Fixing the commons operator initial restarting problem (see https://github.com/stackabletech/commons-operator/issues/111 and spike https://github.com/stackabletech/commons-operator/tree/spike/sts-restarter-webhook)
+
+When creating an API request, e.g. creating a custom resource, the Kubernetes API server performs the following steps before storing into etcd.
```
Techassi, line 23: suggestion: Be slightly more specific how the interception works:

```suggestion
When the Kubernetes API server receives an incoming request, for example when a custom resource should be created, webhooks can intercept that request after authentication and authorization. The following steps are performed before the server finally persists the object into etdc.
```

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+1. Authentication / Authorization
+2. Mutating admission
+3. Schema validation
+4. Validation admission
+5. Conversion to storage version
+
+The steps 2 to 4 may run in a loop if multiple mutating admissions are performed and have side effects.
```
Techassi, line 31: suggestion: There *can* be side effects, but there don't have to be:

```suggestion
The steps 2 to 4 may run in a loop if multiple mutating admissions are performed and can have side effects.
```

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
Techassi: We should be careful to not mix the "admission webhook" concepts with the "CRD versioning webhook" concepts. As far as I know, the versioning webhook lives outside of the flow described below.

Techassi: If we provide explanations for both concepts, we should clearly separate them into different sections.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,211 @@
...
+      caBundle: "Ci0tLS0tQk...<base64-encoded PEM bundle>...tLS0K"
+----
+
+The `clientConfig.caBundle` can not be shipped by us as it will differ from cluster to cluster and must be injected at runtime.
+
+This ADR is about how to achieve the CA bundle injection using external tools or a self-made solution via the secret-operator.
+
+== Problem Statement
+
+The required CA bundles for the webhooks endpoints must be injected at runtime. There exist tools like [Cert Manager]([url](https://cert-manager.io/docs/concepts/ca-injector/#injecting-the-kubernetes-api-server-ca)) that do exactly this, adding one of these annotations to an injectionable source: cert-manager.io/inject-ca-from, cert-manager.io/inject-ca-from-secret, or cert-manager.io/inject-apiserver-ca (depending on the injection source). For example:
```
Techassi, line 132: suggestion: slight reword and listing the different labels:

```suggestion
The required CA bundles for the webhooks endpoints must be injected at runtime. There exist tools like https://cert-manager.io/docs/concepts/ca-injector/#injecting-the-kubernetes-api-server-ca[Cert Manager] that do exactly this.
Adding one of these annotations to an injectionable source:

. `cert-manager.io/inject-ca-from`
. `cert-manager.io/inject-ca-from-secret`
. `cert-manager.io/inject-apiserver-ca`

For example:
```

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+[source,yaml]
+----
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: my-webhook.example.com
+  annotations:
+    cert-manager.io/inject-ca-from-secret: default/my-webhook-example-com-ca
+----
```
Techassi, line 142: note: As I stated above, we need to be careful when mixing both concepts. If we provide this example, we would also need to provide an example for the CRD versioning webhook.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+This can be used for CRDs and conversion webhooks as well.
+
+The SDP should be compatible with as many of these as possible. The normal way of configuration seems to be adding an annotation to the objects containing the CA, which we can probably somehow support in our Helm chart, or solve via documentation.
```
Techassi, line 146: note: I'm actually currently working on supporting settings custom labels when deploying our operators via the Helm chart (using the values.yaml file).

Adding support for annotations via the same mechanism is very similar.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: In general im with you, but i think in the context of CA injection they behave the same?
Step 5 does not have to include a conversion webhook, kubernetes can just stupidly replace the version itself :D

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+== Context
+
+There are many use cases for the future development of the SDP that involve validating, mutating and conversion webhooks. Tasks that must be tackled in the near future include:
+
+- The proper versioning of our CRDs and the possibility of smoothly up- and downgrading between versions including breaking changes (see https://github.com/stackabletech/documentation/issues/273)
+- Fixing the commons operator initial restarting problem (see https://github.com/stackabletech/commons-operator/issues/111 and spike https://github.com/stackabletech/commons-operator/tree/spike/sts-restarter-webhook)
+- Inject logging / vector sidecar containers (mutating)
+
+When the Kubernetes API server receives an incoming request, for example when a custom resource should be created, webhooks can intercept that request after authentication and authorization. The following steps are performed before the server finally persists the object into etc.
```
adwk67, line 24: etcd?
```suggestion
When the Kubernetes API server receives an incoming request, for example when a custom resource should be created, webhooks can intercept that request after authentication and authorization. The following steps are performed before the server finally persists the object into etcd.
```

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,225 @@
...
+  name: my-webhook.example.com
+  annotations:
+    cert-manager.io/inject-ca-from-secret: default/my-webhook-example-com-ca
+----
+
+This can be used for CRDs and conversion webhooks as well.
+
+The SDP should be compatible with as many of these as possible. The normal way of configuration seems to be adding an annotation to the objects containing the CA, which we can probably somehow support in our Helm chart, or solve via documentation.
+
+In a second step the operator needs to obtain a certificate signed by the CA, for which we could use the k8search backend of the secret operator and point this at a configmap which again, can be filled by cert-manager or the Service CA operator mentioned above, configured via an annotation.
```
adwk67, line 156: I find this a little unclear, but that's maybe because I have only looked at `cert-manager.io/inject-apiserver-ca` in detail, and in that case, there is no second step for us to execute: the annotation is picked up and the bundle is injected a few moments later.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: Thanks, was underlined and i just fixed it :D

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
Techassi: Yes, in the context of CA injection, both concepts should use the same mechanisms.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
Techassi: Adding another example might be the better way forward. This way we can show that the CA injection should work with both concepts.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
Techassi: This should be resolved by f5159ce, right?

Techassi: Stating the cert-manager docs:

> NOTE: In this case you will have to ensure that your webhook is configured to serve a TLS certificate that has been signed by the Kubernetes cluster CA. The disadvantages of this mechanism are that: you will require access to the private key of the Kubernetes cluster CA and you will need to manually rotate the webhook certificate.

If we deploy the webhook automatically (which is the current plan afaik), we need to deal with this. Did we look into that?

maltesander: That only applies if you want to use the same CA as the Kubernetes API server.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: I dont follow, what other example do you want? There is one for conversion and validating webhook?

Techassi: Oh sorry, maybe something like:

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  # name must match the spec fields below, and be in the form: <plural>.<group>
  name: crontabs.example.com
  annotations:
    cert-manager.io/inject-ca-from-secret: default/my-webhook-example-com-ca
```

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
Techassi: Yes, that's correct. We should then maybe link (see [here](https://github.com/stackabletech/documentation/blob/f5159cec1fc1108365c1f6802eb4eb85cdbbc210/modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc?plain=1#L133)) to a different section, for example [this](https://cert-manager.io/docs/concepts/ca-injector/#injecting-ca-data-from-a-certificate-resource) or [this](https://cert-manager.io/docs/concepts/ca-injector/#injecting-ca-data-from-a-secret-resource).

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: ah yes, done https://github.com/stackabletech/documentation/pull/522/commits/256812e2ca3119b74b926f08ebd29c67fe78149a

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+The OpenShift Service CA operator is an OpenShift ClusterOperator and contains several controllers:
+
+* *Serving cert signer*: Issues a signed serving certificate/key pair to services annotated with `service.beta.openshift.io/serving-cert-secret-name` via a secret
+* *ConfigMap CA bundle injector*: Watches for configmaps annotated with `service.beta.openshift.io/inject-cabundle=true` and adds or updates a data item (key `service-ca.crt`) containing the PEM-encoded CA signing bundle. Consumers of the configmap can then trust `service-ca.crt` in their TLS client configuration, allowing connections to services that utilize service-serving certificates. Pods referencing the `service-ca.crt` in a VolumeMount will not start before the CA bundle was injected.
+
+[[option3]]
+=== Lightweight self-made solution via secret-operator
+
+This would be the Stackable internal solution to avoid any external party tools. It would work similar to the OpenShift Service CA operator but rather injecting the bundles via the CSI instead of ConfigMap mounts.
```
adwk67, line 189:
```suggestion
This would be the Stackable internal solution to avoid any external party tools. It would work in a similar way to the OpenShift Service CA operator but would rather inject the bundles via the CSI instead of ConfigMap mounts.
```

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,234 @@
...
+=== <<option4>>
+
+* Good, because no external tools are required
+* Good, because all included within the SDP
+* Good, because external tools can be used if required
+* Bad, because requires more time and coding
+
+== Decision Outcome
+
+Chosen option <<option4>>, because the CA injection will be handled as part of the SDP and no external dependencies are required. The common library will reside in the operator-rs and used in every operator. OpenShift should not pose a problem since no extra components are necessary. The required infrastructure (e.g. operator-templating, ca injection / generation) can be developed in parallel to the operators e.g. version conversion logic.
```
adwk67, line 228:
```suggestion
Chosen option <<option4>>, because the CA injection will be handled as part of the SDP and no external dependencies are required. The common library will reside in the operator-rs and be used in every operator. OpenShift should not pose a problem since no extra components are necessary. The required infrastructure (e.g. operator-templating, ca injection / generation) can be developed in parallel to the operators e.g. version conversion logic.
```

adwk67 (dismissed):
lgtm! Just a few minor suggestions.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,234 @@
...
+  # group name to use for REST API: /apis/<group>/<version>
+  group: example.com
+  # list of versions supported by this CustomResourceDefinition
+  versions:
+  - name: v1beta1
+    # Each version can be enabled/disabled by Served flag.
+    served: true
+    # One and only one version must be marked as the storage version.
+    storage: true
+    # Each version can define its own schema when there is no top-level
```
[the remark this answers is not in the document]
maltesander: This does not really make sense to me :) ?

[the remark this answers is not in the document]
adwk67: the rest of the sentence is on the next line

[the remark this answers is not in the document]
adwk67: so: "when no top-level schema is defined"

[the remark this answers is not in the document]
adwk67: or: "when there is no top-level schema defined"

adwk67 (approved):
LGTM!

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+4. Validation admission
+5. Conversion to storage version
+
+The steps 2 to 4 may run in a loop if multiple mutating admissions are performed and can have side effects.
+
+In order for the Kubernetes API server to contact webhook controllers, the webhook endpoints have to present a certificate trusted by the Kubernetes API server. This can be configured similarly for conversion and validating / mutating webhooks as follows.
+
+=== Conversion webhooks CA configuration
+
+[source,yaml]
```
razvan, line 38: Maybe add a short introductory text explaining what the reader is looking at here. What are the important bits ? Can this be reduced to a relevant snippet?

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+    # kind is normally the CamelCased singular type. Your resource manifests use this.
+    kind: CronTab
+    # shortNames allow shorter string to match your resource on the CLI
+    shortNames:
+    - ct
+----
+
+=== Validating / Mutating webhooks CA configuration
+
+[source,yaml]
```
razvan, line 107: Same here.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+----
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: my-webhook.example.com
+  annotations:
+    cert-manager.io/inject-ca-from-secret: default/my-webhook-example-com-ca
+----
+
+This can be used for CRDs and conversion webhooks as well.
```
razvan, line 157: I don't understand the CRDs part. Maybe a short example would help

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+The SDP should focus on one solution for CA injection, internally or externally, but strive for compatibility with as many others as possible. The normal way of configuration seems to be adding an annotation to the objects containing the CA, which can be supported in our Helm chart or solved via documentation.
+
+For clusters without any existing manager we should provide our own, lightweight caBundle injector that can work in tandem with the secret operator.
+
+== Decision Drivers
+
+* Generic solution to be compatible with as many external cert providers as possible
```
razvan, line 165: What does it mean to be compatible with others ? At what level ? Reuse labels ? Crypto algos ?

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+* Openshift compatible
+* How to activate / deactivate if e.g. no conversion webhooks should be applied?
```
razvan, line 167: Activate/deactivate bundle injection? What does this have to do with the question if webhooks should be applied or not ?

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+Put the cert management / injection stuff into a library in operator-rs (similar to the code in secret-operator) that we run in-process for each operator that has a webhook. See https://github.com/stackabletech/commons-operator/tree/spike/sts-restarter-webhook[commons-operator spike].
+
+== Pros and Cons of the Options
+
+=== <<option1>>
+
+* Good, because covers both Kubernetes and Openshift
+* Good, because widely used and de-facto standard for cert injection
+* Good, because cert injection works via annotations (compatibility) for various Kubernetes Resources (CRDs, Validating/Mutating webhooks)
+* Bad, because another tool we have to be experienced with, check for updates and breaking features etc.
```
razvan, line 202: Another major advantage I see is that it's one less (critical) piece of tech that we need to maintain.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+* Good, because no external tools are required
+* Good, because reusing secret-operator and cert management should be an internal part of the SDP
+* Bad, because requires more time and coding
+* Bad, because secret-operator must version its own CRDs as well which could be a predicament
+
+=== <<option4>>
+
+* Good, because no external tools are required
+* Good, because all included within the SDP
```
razvan, line 221: Aren't these two basically saying the same thing? :)

razvan (commented):
my 2€ cents

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,233 @@
...
+This can be used for CRDs and conversion webhooks as well.
+
+The SDP should focus on one solution for CA injection, internally or externally, but strive for compatibility with as many others as possible. The normal way of configuration seems to be adding an annotation to the objects containing the CA, which can be supported in our Helm chart or solved via documentation.
+
+For clusters without any existing manager we should provide our own, lightweight caBundle injector that can work in tandem with the secret operator.
+
+== Decision Drivers
+
+* Generic solution to be compatible with as many external cert providers as possible
+* Openshift compatible
```
razvan, line 166: I forgot to mention that our current CSI based  solution is not OLM compatible. For example, the secret op is a pain to install and impossible to maintain (upgrade).

If we insist in building our own tools, it would be good to have an alternative.

Also, OpenShift now has [it's own secret op](https://docs.openshift.com/container-platform/4.14/storage/container_storage_interface/ephemeral-storage-shared-resource-csi-driver-operator.html)

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: Added a text, i like the full example with the different versions...maybe metadata / groups etc could be removed but no strong opinion here.
https://github.com/stackabletech/documentation/pull/522/commits/00a89cd2092283328567ee5fa015c8be5f6f0adc

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: https://github.com/stackabletech/documentation/pull/522/commits/00a89cd2092283328567ee5fa015c8be5f6f0adc

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: I referenced the example already provided. https://github.com/stackabletech/documentation/pull/522/commits/ffdea691ca4c0c08e8fd898ce6759f364ae668a7

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: tried to clarify https://github.com/stackabletech/documentation/pull/522/commits/d061e00517c4107f3409347542d28fafe060c501

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: The question here is about how to inject. With a solution via cert-manager you will add an annotation for the ca injection. For us this is probably done via helm templating.
If customers do not want any conversion (e.g. always use the storage version and update everything accordingly regarding breaking changes) there must be an option in the templating to not provide the webhook at all (therefore do not add the ca injection annotaiton) as well as not deploying the webhook at all.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: Yeah, this is not quite consistent. The disadvantage of that is mentioned ("Bad, because another tool we have to be experienced with, check for updates and breaking features 
etc.")...

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
maltesander: i would say yes :)

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
razvan: So here is my suggestion:
```suggestion
* Good, because it's already popular and familiar to users
* Good, because it frees maintenance and development resources for us
* Bad, because another tool we have to be experienced with, check for updates and breaking features etc.
```
And I would delete line 200.

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
razvan:
```suggestion
```

razvan:
More on OLM : according to the [documentation](https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html#olm-webhook-considerations_osdk-generating-csvs), OLM does it's own CA injection:

>OLM is configured to provide each deployment with a single certificate authority (CA). The logic that generates and mounts the CA into the deployment was originally used by the API service lifecycle logic. As a result:
> * The TLS certificate file is mounted to the deployment at /apiserver.local.config/certificates/apiserver.crt.
> * The TLS key file is mounted to the deployment at /apiserver.local.config/certificates/apiserver.key.

How will this affect our approach ?

modules/contributor/pages/adr/ADR033-foundation-webhooks-ca-bundle.adoc
```diff
@@ -0,0 +1,253 @@
...
+
+[[option3]]
+=== Lightweight self-made solution via secret-operator
+
+This would be the Stackable internal solution to avoid any external party tools. It would work in a similar way to the OpenShift Service CA operator but would rather inject the bundles via the CSI instead of ConfigMap mounts.
+
+[[option4]]
+=== Common library for cert management / injection in operator-rs
+
+Put the cert management / injection stuff into a library in operator-rs (similar to the code in secret-operator) that we run in-process for each operator that has a webhook. See https://github.com/stackabletech/commons-operator/tree/spike/sts-restarter-webhook[commons-operator spike].
```
sbernauer, line 213: Given this is the chosen option, I think it would be great if you could add one or two additional sentences.
I had a look at the spike branch, but it has 1232 additions and is not self-explaining

Techassi: I created this PR: https://github.com/stackabletech/operator-rs/pull/736 It will implement the auto-generation of TLS certificates. It will of course add a slew of doc comments about the technical details.

Techassi: I should be able to refine this section in the ADR once I have some code in place.

Techassi: This should be addressed in 3277224.

PR #524, opened 2024-01-09
ADR: Conversion webhook deployment

maltesander:
Refinement 2024-02-05:

* Downgrades are impossible on OLM because op and webhook are deployed in lock-step.
* We will extend the ADR to describe the deployment in OLM environment.
* We will extend the ADR to describe the use-cases that we cover with respect to upgrades and downgrades.

preview link: https://deploy-preview-524--stackable-docs.netlify.app/home/nightly/contributor/adr/adr034-foundation-webhooks-deployment


modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
```diff
@@ -0,0 +1,195 @@
...
+==== Pros
+
+- Webhook code could reside in the operator code
+- Operator crash does not affect webhook
+- Could be enabled / disabled Helm parameters
+- Always up to date with the operator, no extra versioning
+
+==== Cons
+
+- No overhead due to operator image (just the lightweight webhook server)
```
adwk67, line 111: this is a "pro" I would say:

> - No overhead due to operator image (just the lightweight webhook server)

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
```diff
@@ -0,0 +1,195 @@
...
+
+- Operator crash does not affect webhook
+- Could be enabled / disabled via custom resource (one webhook per cluster - not sure if that is desired though)
+- Always up to date with the operator, no extra versioning
+- Should not interfere with OpenShift
+
+==== Cons
+
+- Possibly extra image
+- Possibly extra pipelines
```
adwk67, line 130:
```suggestion
- Possibly extra pipelines
- Possibly more complex to test
```

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
maltesander: I would say so too :)
https://github.com/stackabletech/documentation/pull/524/commits/b1f1f929aa65a032d829551ac031b0ae2aaccc74

Techassi (commented):
Good write-up. I prefer option 2 or 3 (or 5).

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
```diff
@@ -0,0 +1,172 @@
...
+** Andrew Kenworthy
+** Malte Sander
+** Sascha Lautenschlaeger
+* Date: {date}
+
+Technical Story: https://github.com/stackabletech/issues/issues/361
+
+// TODO: insert ref to ADR?? CA bundle injection https://github.com/stackabletech/documentation/pull/522
+
+== Context
```
sbernauer, line 18: What would be also awesome context is how k8s does the whole versioning for e.g. Pods and Services to be highly available.

Additionally it would be great to know how other operators solved this problem

sbernauer: Yeah, agreed :+1: However, my comment wasn't about how the versioning works, I would be interested how k8s (or other projects) deploy the stuff highly available

maltesander: I can have a look, but wouldnt this just be multiple pods (which we can with the chosen option) and a service :) ?

sbernauer: Probably yes, but maybe we are also over-engineering and they just have a single operator that exposes the HTTP endpoint. Who knows :D
So yeah, would be nice context, but definitively no blocker for merging - so feel free to ignore :sweat_smile:

maltesander: Yeah will have a look for the other ADR. Just to comment: I think we are far better of with separating this functionality for each operator. I did not see any examples where CRDs are handled by the same webhook (which is certainly possible). This brings us back in versioning hell again as well :D

fhennig:
Can you link this in the epic? It's this one right? https://github.com/stackabletech/issues/issues/361

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
```diff
@@ -0,0 +1,176 @@
...
+Additionally, webhooks should be enabled / disabled on demand via options like Helm, operator-parameters or CRD flags.
+
+== Decision Drivers
+
+- Do we want this to be HA?
+- Do we want this to be deployed decoupled?
+- One operator per kubernetes cluster: What if 3 operators deployed watching different namespaces / versions? Should be strongly discouraged!
+- How to abstract a common admission/conversion webhook skeleton in operator-rs, that can be implemented in the operators within a few lines of code (excluding the actual conversion code)?
+- How to keep maintenance, updating, pipelines or extra images to a minimum?
+- Openshift compatability?
```
fhennig, line 52:
```suggestion
- Openshift compatibility?
```

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+[[option5]]
+=== Option 5: The webhook is a deployment via Helm
+
+The operator Helm bundle contains an extra dedicated deployment with the webhook server (in the operator namespace?). One webhook deployment per operator. Same image as operator.
+
+==== Pros
+
+- Operator crash does not affect webhook
+- Downgrade possible -> can adept to new CRD storage versions
```
fhennig, line 145: It's not entirely clear to me how downgrading would work. Would I uninstall the operator and then install the previous version of it, but keep the webhook image of the new version already? Seems a bit messy, it sounds like maybe 2 Helm Charts would be better in that case? But maybe I just got it wrong!

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
```diff
@@ -0,0 +1,176 @@
...
+== Problem Statement
+
+There are several options on how or where to deploy a conversion webhook, e.g. coupled closely with the operator as a controller or completely decoupled via an extra helm deployment.
+
+We need a uniform deployment across all operators to keep implementation and maintenance to a minimum and reuse code wherever possible.
+Additionally, webhooks should be enabled / disabled on demand via options like Helm, operator-parameters or CRD flags.
+
+== Decision Drivers
+
+- Do we want this to be HA?
```
razvan, line 47: For future readers, we should probably put these criteria in terms of affirmative statemens like:
```
* must be HA.
* ...
```
and not leave them as open questions.

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
```diff
@@ -0,0 +1,176 @@
...
+== Considered Options
+
+[[option1]]
+=== Option 1: Deploy within the operator as controller
+
+The operator contains another controller / thread with the webhook server and conversion code.
+
+==== Pros
+
+- No extra bin / main file
```
razvan, line 65: This is not part of the decision drivers list. It's good that we think about it but it doesn't feel like it belongs in the ADR. The same applies for next one.

razvan (commented):
my comments

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
maltesander: Lets say im upgrading, the CRD contains a new version (which then e.g. is the storage version as well) and replaces the old CRD. Then you can just install the older operator (e.g. without replacing the CRD) because the older version is still there. The webhook will ideally be always the "latest", since it has all bugfixes, all known / supported versions. If this is not the case, e.g. the new version will not be understood by an older webhook.

So yeah, i think we should have this independent (which is not excluded in that option). And since the webhook will be its own deployment it will have its own image which could "just" pull in latest etc. (please dont quote me on this!).

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
maltesander: I would say its included in?
```
- How to keep maintenance, updating, pipelines or extra images to a minimum?
```

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
maltesander: Yeah that is true i try to improve that.

maltesander: https://github.com/stackabletech/documentation/pull/524/commits/e04c7fc99f909ea073986fde02b0b6621e65e8eb

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
razvan: IMO this needs more refinement.
* What about OpenShift OLM ? Users cannot choose to install the operator but not the webhook. The operator package always contains both.
* What does a downgrade mean after a full install look like? Helm would uninstall both (i guess). Then users would need to reinstall the new webhook and install the old operator in two different steps. Is this correct?
* What is the default installation? Both ? just the operator ?

modules/contributor/pages/adr/ADR034-foundation-webhooks-crd-versioning.adoc
razvan: I know it's somewhat important to us but it's not important to the users. I would say for the scope of the ADR it's irrelevant.

razvan:
More on OpenShift/OLM: according to the [documentation](https://docs.openshift.com/container-platform/4.14/operators/operator_sdk/osdk-generating-csvs.html#olm-webhook-considerations_osdk-generating-csvs) there are some constraints on conversion webhooks. The relevant to is:

>OLM places the CSV in the failed phase if a conversion webhook definition does not adhere to the following constraints:
> * CSVs featuring a conversion webhook can only support the AllNamespaces install mode.

We currently do not support it because we our operators can only be installed in one namespace (`stackable-operators`). This restriction comes from the way we deploy the secret operator and we may need to change it.

razvan (dismissed):

Techassi:
I'm still working on this. I will add some smaller adjustments.

Techassi (approved):

modules/contributor/pages/adr/ADR034-foundation-webhooks-deployment.adoc
```diff
@@ -0,0 +1,234 @@
...
+
+== Context
+
+We must version our CustomResourceDefinitions (CRDs).
+This step allows us to move away from unstable alpha or beta versions (like `v1alhpa1`) to stable versions like `v1` or `v2`.
+These versions provide stable interfaces which customers can rely on.
+Since we cannot avoid having breaking changes in the future (which require a bump in the respective CRD version), we have to supply conversion webhooks that take care of converting older versions to the current storage version.
+
+Converting custom resources between versions is a separate step, independent of webhook deployments.
+CRD versions should be seemlessly upgraded when new operators/webhooks are upgraded. Downgrades are possible by first converting the custom resources to the old version and then downgrading the operator and webhook.
```
adwk67, line 27:
```suggestion
CRD versions should be seamlessly upgraded when new operators/webhooks are upgraded. Downgrades are possible by first converting the custom resources to the old version and then downgrading the operator and webhook.
```

adwk67 (approved):
lgtm (just one typo)

adwk67 (approved):

PR #533, opened 2024-01-23
Fix deprecated strings

nightkr:


sbernauer (approved):

sbernauer:
@nightkr just FYI you might want to change them in the releases branches as well

PR #534, opened 2024-01-26
Remove floating tag references in custom image selection

maltesander:


sbernauer (approved):
Floating tags, we will miss you :cry:

PR #535, opened 2024-01-29
docs: Add Cargo.toml style guide

### LINKED ISSUE #528: Document Cargo.toml formatting
It is good to have consistent formatting for consistency between developers (eg: using rustfmt for ust code) to reduce noise around style in PR reviews.

Automatic formatting is a preference, however exising TOML formatters (as far as we know) apply rules across the whole toml file (and not only Cargo.toml), rather than different rules per section or use case of toml.

For example,
- we would like `[dependencies]` to be ordered alphabetically
- we would like important `[project]` attributes to be at the top of the section

So, for now we should document how we expect Cargo.toml to look, to avoid it repeatedly coming up in PRs.

fhennig:
fixes #528

### DIFF
--- a/modules/contributor/pages/code-style-guide.adoc
+++ b/modules/contributor/pages/code-style-guide.adoc
@@ -199,7 +199,7 @@ enum Colors {
 
 === Formatting of struct fields and enum variants
 
-We add newlines to struct fields and enum variants when they include additional information like documentation comments or attributes, because the variants can become difficult to read.
+Add newlines to struct fields and enum variants when they include additional information like documentation comments or attributes, because the variants can become difficult to read.
 This is especially the case when fields include doc comments, attributes like `#[snafu()]`, and in case of enum variants, various embedded types.
 
 Enum variants and struct fields don't need to be separated when **no** additional information is attached to any of the variants or fields.
@@ -279,12 +279,12 @@ Comments should only be added if they provide additional information not availab
 
 === Choice of error crate and usage
 
-We use `snafu` for all error handling in library *and* application code because we want to provide as much context to the user as possible.
+Use `snafu` for all error handling in library *and* application code to provide as much context to the user as possible.
 Further, `snafu` allows us to use the same source error in multiple error variants.
-This feature can be used for cases were we need / require more fine-grained error variants.
+This feature can be used for cases where more fine-grained error variants are required.
 This behaviour is not possible when using `thiserror`, as it uses the `From` trait to convert the source error into an error variant.
 
-Additionally, we restrict the usage of the `#[snafu(context(false))]` atrribute on error variants.
+Additionally, the usage of the `#[snafu(context(false))]` atrribute on error variants is restricted.
 This ensures that fallible functions need to call `.context()` to pass the error along.
 
 The usage of `thiserror` is considered invalid.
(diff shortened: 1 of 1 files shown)


modules/contributor/pages/code-style-guide.adoc
```diff
@@ -1,5 +1,55 @@
 = Source code style guide
 
+== Cargo.toml
+
+Follow the https://doc.rust-lang.org/nightly/style-guide/cargo.html[official formatting conventions] for the `Cargo.toml` file.
```
Techassi, line 5: suggestion: Use we.

```suggestion
We follow the https://doc.rust-lang.org/nightly/style-guide/cargo.html[official formatting conventions] for the `Cargo.toml` file.
```

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -1,5 +1,55 @@
...
+This means:
+
+* Put the `[package]` section at the top of the file
+* put the name and version keys in that order at the top of that section, followed by the remaining keys other than description in order, followed by the description at the end of that section.
```
Techassi, line 9: note: Capitalize the letter `p`.

Techassi, line 9: note: What order are we talking about here? The same order as mentioned below? If so, we should link to it here already.

Techassi (changes requested):

modules/contributor/pages/code-style-guide.adoc
fhennig: Actually, we shouldn't use "we" in our docs: https://docs.stackable.tech/home/stable/contributor/style-guide#_tone_and_writing_style_google_developer_documentation_style_guide

This is instructional "Do it like this"

Techassi: Alright, then we should probably change that for the rest of the guide as well.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -1,5 +1,55 @@
 = Source code style guide
 
+== Cargo.toml
+
+Follow the https://doc.rust-lang.org/nightly/style-guide/cargo.html[official formatting conventions] for the `Cargo.toml` file.
+This means:
+
+* Put the `[package]` section at the top of the file.
+* Put the name and version keys in that order at the top of that section, followed by the remaining keys other than description in order (sort keys with https://www.gnu.org/software/coreutils/manual/html_node/Version-sort-overview.html[version-sort]; very similar to lexical sorting)), followed by the description at the end of that section.
```
Techassi, line 9: suggestion: Use inline code blocks.

```suggestion
* Put the `name` and `version` keys in that order at the top of that section, followed by the remaining keys other than `description` in order (sort keys with https://www.gnu.org/software/coreutils/manual/html_node/Version-sort-overview.html[version-sort]; very similar to lexical sorting)), followed by the `description` at the end of that section.
```

modules/contributor/pages/code-style-guide.adoc
fhennig: probably, but maybe not in this PR

fhennig: or maybe I can do it :grin:

fhennig: done

Techassi (approved):

PR #545, opened 2024-02-08
Add a section on Export Controls

lfrancke:
The rendered page can be found here: https://deploy-preview-545--stackable-docs.netlify.app/home/nightly/export

@fhennig The link in the navbar is now dead for 23.11 and previous releases, do we need to back port it somehow or what would you suggest?

### DIFF
--- a/modules/ROOT/pages/export.adoc
+++ b/modules/ROOT/pages/export.adoc
@@ -6,7 +6,7 @@ The USA adopts the https://en.wikipedia.org/wiki/Export_Administration_Regulatio
 
 All of our products are outside of the scope of EAR because they fall under the _publicly available_ exemption and do not contain non-standard cryptography.
 
-NOTE: That means E-Mail notifications to the NSA and BIS are not required and therefore not published here.
+NOTE: That means E-Mail notifications to the NSA and BIS are not required and are therefore not published here.
 
 In particular:
 
--- a/modules/ROOT/pages/export.adoc
+++ b/modules/ROOT/pages/export.adoc
@@ -9,7 +9,7 @@ That means E-Mail notifications to the NSA and BIS are not required and therefor
 
 In particular:
 
-* We are exempt under https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.3[EAR 734.3(b)] because we are published according to EAR 734.7
+* We are exempt under https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.3[EAR 734.3(b)] because we are published according to https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.7[EAR 734.7]
 * We are exempt under https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.7[EAR 742.15(b)] as all our software only includes https://ecfr.io/Title-15/Section-772.1[standard encryption]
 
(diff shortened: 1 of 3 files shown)


lfrancke:
Requested review by lawyer toda.

fhennig:
> The link in the navbar is now dead for 23.11 and previous releases, do we need to back port it somehow or what would you suggest?

Oh, good point, this is actually also the case for the other pages in that menu. Yes I believe backporting them would be the easiest in this case. It also makes sense since the information applies to all releases.

lfrancke:
Thank you for the offer.

Yes, please do put it on the board and work on it at when you have the time.

fhennig:
https://github.com/stackabletech/documentation/issues/547 done

modules/ROOT/pages/export.adoc
```diff
@@ -0,0 +1,16 @@
+= Export Control
+
+== USA
+
+The USA have the https://en.wikipedia.org/wiki/Export_Administration_Regulations[Export Administration Regulations (EAR)] as the primary regulation to control export.
```
adwk67, line 5:
```suggestion
The USA adopts the https://en.wikipedia.org/wiki/Export_Administration_Regulations[Export Administration Regulations (EAR)] as the primary regulation to control exports.
```

modules/ROOT/pages/export.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+All of our products are outside of the scope of EAR because they fall under the _publicly available_ exemption and do not contain non-standard cryptography.
+
+NOTE: That means E-Mail notifications to the NSA and BIS are not required and therefore not published here.
```
adwk67, line 9:
```suggestion
NOTE: That means E-Mail notifications to the NSA and BIS are not required and are therefore not published here.
```

modules/ROOT/pages/export.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+In particular:
+
+* We are exempt under https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.3[EAR 734.3(b)] because we are published according to https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.7[EAR 734.7]
```
adwk67, line 13:
```suggestion
* We are exempt under https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.3[EAR 734.3(b)] because we publish according to https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.7[EAR 734.7]
```

modules/ROOT/pages/export.adoc
```diff
@@ -0,0 +1,16 @@
...
+* We are exempt under https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-734/section-734.7[EAR 742.15(b)] as all our software only includes https://ecfr.io/Title-15/Section-772.1[standard encryption]
```
adwk67, line 14:
```suggestion
* We are exempt under https://www.ecfr.gov/current/title-15/subtitle-B/chapter-VII/subchapter-C/part-742/section-742.15[EAR 742.15(b)] as our software includes only https://ecfr.io/Title-15/Section-772.1[standard encryption]
```

adwk67 (changes requested):

modules/ROOT/pages/export.adoc
lfrancke: Thanks for catching this

lfrancke:
Thanks for the review and the suggestions

adwk67 (approved):
LGTM

PR #546, opened 2024-02-08
Updated dependencies

lfrancke:


fhennig:
Why was the checkout action downgraded from 3.2.0 to 3.1.1?

lfrancke:
Good catch! It is actually 4.1.1 not 3.1.1. I'll fix it!

PR #559, opened 2024-02-28
Adding warning and hint for listener operator

Maleware:
Added warning for listener operator not working in 23.11.0

I've installed the ./arm.sh script on Azure and got an image pull backoff. Research showed that the pods

`k8s_external-provisioner_listener-operator-controller-deployment` and `k8s_node-driver-registrar_listener-operator-node-daemonset` are still `amd64`.


sbernauer (approved):
LGTM. However, wouldn't it be easier to ship listener-op for arm as well? It is required for HDFS since ~1 month and the need for it will increase in the future.
I recall that some compilation problem was fixed by pulling the git submodule

PR #561, opened 2024-03-11
docs: Add webhook server implementation guide / tutorial

Techassi:
Fixes https://github.com/stackabletech/issues/issues/540

Preview: https://deploy-preview-561--stackable-docs.netlify.app/home/nightly/contributor/guidelines/webhook-server

PR #562, opened 2024-03-12
chore: Add page aliases

Techassi:
Follow-up of #561


fhennig (approved):

Techassi:
Ohhh, okay :)

Techassi:
Yeah... It feels a little pointless. To work like we expected, it would need an `index.html` file in the `logging` folder which then redirects to `guidelines/logging` (which again is an `index.html` file).

PR #574, opened 2024-03-26
feat: Add the Atlas extension to allow for partial builds

fhennig:
The extension generates a `https://docs.example.org/site-manifest.json` file, nothing more, so this change does not affect the docs itself.

We can use this site manifest down the road for partial builds, allowing faster dev cycles and possibly docs previews in the operator repositories.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -188,7 +188,11 @@ ui:
   supplemental_files: ./supplemental-ui
 antora:
   extensions:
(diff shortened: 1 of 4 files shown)


razvan (approved):
ok

fhennig:
Couldn't get partrial builds to work, shelved for now.

The extension worked, and a site-manifest was generated, but partial builds did not work. Maybe it'll get stabilized with 3.2.0

PR #579, opened 2024-03-28
Add playbook to only build dev version

sbernauer:
To speed up feedback loop


soenkeliebau:
Can we pull out the branches into a var '[main, 23.11, ...]'and override that with '[main]' for the dev only run, instead of duplicating the entire playbook?

sbernauer:
What var are you referring to? yaml anchors or is this an Ansible playbook? Looking at https://docs.antora.org/antora/latest/playbook/ I can't find any concept of variables.

I think I would prefer yaml anchors anyway and we could do some 'yq'' magic in the Makefile

fhennig:
It's possible to specify `content.branches` directly: https://docs.antora.org/antora/latest/playbook/configure-content-sources/

So you don#t have to configure branches per source. 

What do you mean with a var though? @soenkeliebau

soenkeliebau:
> What do you mean with a var though? @soenkeliebau

I was being stupid and mistook this for Ansible :(

Makefile
```diff
@@ -1,7 +1,7 @@
 CURRENT_COMMIT := $(shell git rev-parse HEAD)
 
 build-dev: build-ui
-	node_modules/.bin/antora generate local-antora-playbook.yml
+	node_modules/.bin/antora generate only-dev-antora-playbook.yml
```
fhennig, line 4: I would like to have a make step for local and for only-dev. And maybe the makefile steps should have the same name as the playbooks. So `make build-only-dev` and `make build-local`?

local-antora-playbook.yml
```diff
@@ -13,7 +13,7 @@ urls:
 content:
   sources:
     - url: ./
-      branches:
+      branches: &branches
```
fhennig, line 16: I would like to keep this file as it is for a few reasons. There are scripts that rely on the structure of the file being the way it is (part of the release process).

only-dev-antora-playbook.yml
```diff
@@ -0,0 +1,94 @@
...
+content:
+  sources:
+    - url: ./
+      branches:
+        - HEAD
+    # management tools
+    - url: https://github.com/stackabletech/stackable-cockpit.git
+      start_path: docs
+      branches: &branches
+        - main
```
fhennig, line 22: Not really sure how useful the anchor is here, I feel like it just makes it more complicated without really saving anything.

fhennig (approved):

fhennig (changes requested):
see above

Makefile
```diff
@@ -1,6 +1,9 @@
 CURRENT_COMMIT := $(shell git rev-parse HEAD)
 
-build-dev: build-ui
+build-only-dev: build-ui
+	node_modules/.bin/antora generate only-dev-antora-playbook.yml
+
+build-local: build-ui
 	node_modules/.bin/antora generate local-antora-playbook.yml
```
fhennig, line 7:
```suggestion
build-local: build-ui
	node_modules/.bin/antora generate local-antora-playbook.yml

build-only-dev: build-ui
	node_modules/.bin/antora generate only-dev-antora-playbook.yml
```

either the default recipe needs to be the `local` one, or the `.github/build.yml` needs to be updated to call the `make build-local` explicitly.

README.adoc:25 also just mentions "make", also would need to be updated. 

I'm not sure whether we want the local or only-dev one to be the default. I think i#d like "only-dev" to be the special case and local the default, hmm. Not sure.

thanks for all the other changes, look great!

sbernauer: Haha, I though of defaulting to only the dev versions by default.
But I don't have a strong opinion, so picking yours :)

fhennig (approved):
Alright, thanks!

maybe we can change the default in the future :)

sbernauer:
Thanks for the review!

PR #593, opened 2024-04-23
feat: new menu UI and new search

### LINKED ISSUE #583: Improve search by using Algolia

fhennig:
closes #583


Techassi (approved):

lfrancke:
out of curiosity: Is this using an external service for search?

fhennig:
No it isn't. I experimented with that, but a big drawback is that the external calls would require more privacy notices etc., so this is all purely static. The search generates an index that is hosted alongside of the website, and the search uses the self-hosted index, all calls are just site-internal. This also means that there is no crawling required, and the search is always up-to-date, as the index is built every time the page is deployed. Oh, and [pagefind](https://pagefind.app/) is written in Rust too :smile:

lfrancke:
Awesome, thank you!

PR #622, opened 2024-06-18
Fix typo German instead of English

PaulienVa:


lfrancke (approved):

lfrancke:
Thank you for catching this!

PaulienVa:
No pb!

lfrancke:
This is against the 24.3 release branch - we'll also get it fixed on main

Ping @fhennig

lfrancke:
And Felix, please also check why the build fails. Is it because it comes from an external fork?

fhennig:
The build is failing because it isn't running on the main branch. Only the playbook from the main branch can be used to build the documentation.

But thanks for the contribution Paulien! I will move the  change to the main branch as well :)

PR #630, opened 2024-07-17
Use Nexus instead of Harbor in airgap docs

dervoeti:
The docs currently mention Harbor, which works but is not yet officially supported by us. Thus, setting up the mirror inside the airgap registry following the current docs won't have any useful effect since Kubernetes will try to pull all images from `docker.stackable.tech` (Nexus) instead of `oci.stackable.tech`.
I changed the documentation to use Nexus instead. We can just revert this once we switch everything to Harbor.


razvan (approved):
lgtm

PR #634, opened 2024-07-30
docs(release-notes): Add retroactive mention of User Info Fetcher being experimental

NickLarsenNZ:
Add a note to the release notes about UIF being experimental (since it was announced in 24.3)

### DIFF
--- a/modules/ROOT/pages/release-notes.adoc
+++ b/modules/ROOT/pages/release-notes.adoc
@@ -103,7 +103,8 @@ Documentation::
 
 * Apache Spark: documentation has been added showing how to provision Spark dependencies
 
-* Open Policy Agent: N.B. As mentioned in the release 24.3, we will be actively building out the backends supported by the User Info Fetcher. This feature should be therefore be treated as experimental, as implementation details are subject to change as we consolidate and fine-tune the tool.
+* Open Policy Agent: N.B. As mentioned in the release 24.3, we will be actively building out the backends supported by the User Info Fetcher.
+This feature should be therefore be treated as experimental as we continue to extend and consolidate back-end handling and fine-tune the tool in general.
 
 Other product features::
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/release-notes.adoc
```diff
@@ -103,6 +103,8 @@ Documentation::
 
 * Apache Spark: documentation has been added showing how to provision Spark dependencies
 
+* Open Policy Agent: N.B. As mentioned in the release 24.3, we will be actively building out the backends supported by the User Info Fetcher. This feature should be therefore be treated as experimental, as implementation details are subject to change as we consolidate and fine-tune the tool.
```
nightkr, line 106: "Implementation details" is a bit too narrow. We're always free to change implementation details, the problem is that there is no public API yet...

adwk67: how about "usage details" or is that too clunky?

nightkr: Feels clunky but it could work.

adwk67: Or just:

> This feature should be therefore be treated as experimental as we continue to extend and consolidate back-end handling and fine-tune the tool in general.

nightkr: :+1:

adwk67:
```suggestion
* Open Policy Agent: N.B. As mentioned in the release 24.3, we will be actively building out the backends supported by the User Info Fetcher.
This feature should be therefore be treated as experimental as we continue to extend and consolidate back-end handling and fine-tune the tool in general.
```

adwk67 (approved):

adwk67 (approved):

Issue #637, opened 2024-08-01
fix: remove `home:` from refs


NickLarsenNZ:
> From @fhennig:
> The problem is in the include directives in the operators, on the old branches. The [docs](https://docs.antora.org/antora/latest/page/include-a-page/) state:
>> If the version coordinate isn’t specified, Antora uses the latest version of the target page’s component to complete the resource ID at runtime. This behavior only applies when the target page and current page belong to different docs components.
>
> This means that if a component is specified in an include directive (i.e. "home") but without a version, the latest version is used. We do not want that we want to include from the same version. The fix is to simply remove the component directive. I've made a fix for Airflow 24.3, here is the commit: https://github.com/stackabletech/airflow-operator/commit/669691c5a303f5d61ce44e4672a476d79e2208ea
>
> unfortunately this means that you will need to push this fix into all release branches of all operators ... so a lot of PRs

Example fix:

```diff
- include::home:concepts:stackable_resource_requests.adoc[]
+ include::concepts:stackable_resource_requests.adoc[]
```

1. Checkout the `release-xx.xx` branch (make sure it is up to date)
2. Create a new branch off it, `fix/release-xx.xx-remove-home-xref`
   Sample commit message: `docs: remove home module from refs (stackabletech/documentation#637)` 
3. Raise a PR back into the `release-xx.xx` branch

> [!NOTE]
> The documentation repo uses the `release/xx.xx` style.

Update the below tasklists with the subsequent PR

## Non-Operators Repos

```[tasklist]
### Fixes for documentation releases
- [ ] https://github.com/stackabletech/documentation/pull/638
- [ ] https://github.com/stackabletech/documentation/pull/640
- [ ] https://github.com/stackabletech/documentation/pull/641
- [ ] https://github.com/stackabletech/documentation/pull/642
- [ ] https://github.com/stackabletech/documentation/pull/643
- [x] nothing in 24.7
```

```[tasklist]
### Fixes for demos releases
- [ ] https://github.com/stackabletech/demos/pull/76
- [ ] https://github.com/stackabletech/demos/pull/77
- [ ] https://github.com/stackabletech/demos/pull/79
- [ ] https://github.com/stackabletech/demos/pull/78
- [ ] https://github.com/stackabletech/demos/pull/80
- [ ] https://github.com/stackabletech/demos/pull/81
- [ ] https://github.com/stackabletech/demos/pull/82
- [ ] https://github.com/stackabletech/demos/pull/83
[...]

PR #644, opened 2024-08-05
docs: document some antora build details

fhennig:
During the last release we encountered some issues involving xrefs, I have now updated the troubleshooting page and docs overview explanation to make the component structure and xref mechanism more clear.

related to: https://github.com/stackabletech/documentation/issues/637

### DIFF
--- a/modules/contributor/pages/docs/overview.adoc
+++ b/modules/contributor/pages/docs/overview.adoc
@@ -46,7 +46,7 @@ The guide has to account for different use-cases (i.e. the user is using their o
 Since this kind of information is typically product specific, it is located in the usage guide section of individual operators.
 
 **Reference** information for the Stackable platform entails all the settings and Options in our YAMLs, which we generate.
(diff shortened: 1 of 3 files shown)


NickLarsenNZ (approved):
Thanks @fhennig

I also wonder if we can mimic the release branch process before the release (and before all of the dependent release branches exist)? Maybe it would catch the recent issues earlier.

fhennig:
@NickLarsenNZ Yeah that would be nice. The issue happened because there were mistaken links from older versions onto stable. I think we purged them all now, so the issue cannot happen again. I'm more worried about new xrefs sneaking in that also reference 'home' explicitly. 

The process to check for this in advance would be maybe to set `pre-release` to `false` in `nightly`and see if anything breaks, but withouth publishing this change. I find it difficult to come up with a rigid process to do this, I'd lean towards not doing anything for now because of this. I feel like any sort of advance check would introduce too many new complexities that make it not worth it.

PR #665, opened 2024-09-16
Add link checking to detect broken links

fhennig:
WIP


fhennig:
This was a test. The Plugin takes wayyy too long to run it on every build. Not a good solution.

PR #667, opened 2024-09-19
docs(code-style-guide): Add unit test function name rule

Techassi:
This PR adds a rule for naming unit test functions.

### DIFF
--- a/modules/contributor/pages/code-style-guide.adoc
+++ b/modules/contributor/pages/code-style-guide.adoc
@@ -296,7 +296,7 @@ The usage of `thiserror` is considered invalid.
 ----
 #[derive(Snafu)]
 enum Error {
-    #[snafu(display("failed to read config file of user {user_name}"))]
+    #[snafu(display("failed to read config file of user {user_name:?}"))]
     FileRead {
         source: std::io::Error,
         user_name: String,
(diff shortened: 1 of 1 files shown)


modules/contributor/pages/code-style-guide.adoc
```diff
@@ -595,3 +595,57 @@ let memory: CpuQuantity = "2".parse();
...
 
 ====
+
+== Writing tests
+
+=== Unit test function names
+
+Function names of unit tests must not include the `test_` prefix, because otherwise `cargo test` will produce superfluous output like `memory::test::test_addition::case_1`.
+Instead, use an appropriate name to describe what is being tested.
+The previously mentioned name could for example be renamed to `memory::test::addition::case_1`.
```
NickLarsenNZ, line 605: I've split this paragraph into:
- Rule
- Why
- Action

```suggestion
Function names of unit tests must not include a redundant `test` prefix or
suffix.

It results in the output of `cargo test` containing superfluous mentions of
"test", especially when the containing module is called `test`. For example:
`my_crate::test::test_valid`.

Instead, use an appropriate name to describe what is being tested. The previous
example could then become: `my_crate::test::parse_valid_api_version`.
```

I added blank lines for readability, but you're welcome to change that in the suggestion. I also change the example test to line up with the ones in the code examples below.

NickLarsenNZ (approved):
LGTM, made a suggestion on the paragraph.

PR #668, opened 2024-09-26
docs(code-style-guide): Add unit test function name guidance

Techassi:
This adds a bunch of rules regarding Rust comments.

### DIFF
--- a/modules/contributor/pages/code-style-guide.adoc
+++ b/modules/contributor/pages/code-style-guide.adoc
@@ -275,6 +275,88 @@ If any of the items are commented, all items should be.
 It should however also be noted that there is no requirement to comment fields or variants.
 Comments should only be added if they provide additional information not available from context.
 
+== Comments
+
+General rules apply for both doc comments and developer comments.
+The comments should not exceed a line-width of 100 characters with the exception of URLs.
+Long inline URLs should be avoided.
+Put links onto their own line or reference them with Markdown references instead.
+Comments should always form complete sentences with full stops at the end.
+
+[TIP.code-rule,caption=Examples of correct code for this rule]
+====
+
+[source]
+----
+                    100 characters v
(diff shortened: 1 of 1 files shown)


sbernauer (commented):
Works for me. Do we need a decision for this? (e.g. I personally prefer 120 chars, but could live with 100)

Techassi:
> Works for me. Do we need a decision for this? (e.g. I personally prefer 120 chars, but could live with 100)

We can open a decision for this, but tbh, I wouldn't push beyond 100 characters. Rust's `std` and `core` library for example also use line-widths between 80 and 100.

sbernauer (approved):
Works for me. I think we should put it on the decision board, just so that people could veto - but I don't expect any veto

Techassi:
Moving this into voting. Options are:

- :+1: Accept as is
- :-1: Don't accept, post desired changes in comments

stefanigel:
"line-widths between 80 and 100" - I was not aware that Rust is so close to FORTRAN :-)

Techassi:
> "line-widths between 80 and 100" - I was not aware that Rust is so close to FORTRAN :-)

I should clarify: Just for the comments, code is usually formatted at 100 and above. But yes, developers like their narrow comments :)

NickLarsenNZ:
I just like narrow for reviewing side by side. Otherwise I like 100.

No action needed, just sharing.

razvan:
Does the max line length apply to code too?

Techassi:
> Does the max line length apply to code too?

No, we are only talking about comments. Rust code is formatted by `rustfmt` anyway and it will format with a width of ~100 and above.

Techassi:
There seems to be no further activity. I will mark this decision as accepted tomorrow and move it to "Done" afterwards.

PR #672, opened 2024-10-15
Update antora.yml to mark 23.7 as EoL

xeniape:


xeniape:
The error can probably be ignored, that's a result of https://docs.antora.org/antora/latest/page/include-a-page/#module-or-component (third bulletpoint) since we are using HEAD/main branches in the antora-playbooks and they are marked as pre-release, so it's taking referenced versions of components from the first release version from the list in the playbook, which is 23.7 in this case resulting in dublicates. This problem shouldn't affect anything since we are not building the documentation from this branch, but from the main branch.

sbernauer (approved):

PR #674, opened 2024-10-22
chore: Update the footer email and copyright year

NickLarsenNZ:
It would be nice to make the year dynamic. We could add a bit of JS in the header to update the element.


NickLarsenNZ:
Correct PR made here: https://github.com/stackabletech/documentation-ui/pull/77

PR #683, opened 2024-11-26
Add docs on how to mark StorageClass as default in AWS

sbernauer:
Ran into this preparing a customer demo on EKS using replicated


adwk67 (approved):
LGTM

PR #687, opened 2024-11-29
fix: playbook references

NickLarsenNZ:
Thanks @xeniape and @razvan for spotting this.

Relevant [Slack discussion](https://stackable-workspace.slack.com/archives/C02LP2HLM9Q/p1732889838952629).

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -26,6 +26,7 @@ content:
     - url: .
       branches:
         - HEAD
+        - release/24.11
         - release/24.7
         - release/24.3
         - release/23.11
@@ -41,6 +42,7 @@ content:
       start_path: docs
       branches:
         - main
+        - release-24.11
         - release-24.7
         - release-24.3
         - release-23.11
@@ -52,6 +54,7 @@ content:
       start_path: docs
       branches:
         - main
+        - release-24.11
         - release-24.7
         - release-24.3
         - release-23.11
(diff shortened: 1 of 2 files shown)


razvan (approved):
lgtm

antora-playbook.yml
```diff
@@ -26,6 +26,7 @@ content:
     - url: .
       branches:
         - HEAD
+        - release/24.11
```
NickLarsenNZ, line 29: Blocking the merge. This might be wrong.

Previous branches don't include themselves as they are covered by HEAD (even though the site is deployed from main).

NickLarsenNZ:
Closing this, as it doesn't seem to be the fix

NickLarsenNZ:
- The link to <https://docs.stackable.tech/home/24.11/concepts/overrides/> is currently invalid, and only accessible via <https://docs.stackable.tech/home/stable/concepts/overrides/>. After 25.3, those links will work again, but be broken in the same way for 25.3.

`stable` is set here: <https://github.com/stackabletech/documentation/blob/8b307104cc19f3b75c2ed0426794f10d56f4a3d6/antora-playbook.yml#L20C27-L20C33>, maybe there is some way to allow the latest version to also be allowed (rewritten or redirected).

Techassi:
I researched why `latest_version_segment_strategy` doesn't redirect like we expect. The behaviour of this option is influenced by the `redirect_facility` option. I'll have to research if there is a combination of options which work for our deployment.

PR #697, opened 2025-01-02
Add concepts page on JVM argument overrides

sbernauer:


modules/concepts/pages/overrides.adoc
```diff
@@ -150,3 +150,73 @@ The `podOverrides` will be merged onto the following resources the operators dep
...
 * Jobs, that are used to setup systems the product depends on e.g. create a database schema for Superset or Airflow.
+
+[#jvm-argument-overrides]
+== JVM argument overrides
+
+You can configure the JVM arguments used by JVM based tools.
+This is often times needed to e.g. configure a HTTP proxy or other network settings.
+
+As with other overrides, the operator generates a set of JVM arguments that are needed to run the tool, you can specify additional arguments, which are merged on top of the ones the operator generated.
+As some JVm arguments are mutually exclusive (think of `-Xmx123m` and `-Xmx456m`), you also have the option to remove JVM arguments - either by specifying the exact argument or a regex.
```
maltesander, line 161:
```suggestion
As some JVM arguments are mutually exclusive (think of `-Xmx123m` and `-Xmx456m`), you also have the option to remove JVM arguments - either by specifying the exact argument or a regex.
```

modules/concepts/pages/overrides.adoc
```diff
@@ -150,3 +150,73 @@ The `podOverrides` will be merged onto the following resources the operators dep
...
+
+The merging mechanism is applied <operator-generated> <- <role user specified> <- <rolegroup user specified> and works as following:
+
+Let user be the user-specified arguments and operator be the arguments generated by the operator.
```
maltesander, line 165: The `user-specified` is inconsistent with the `user specified` above. I think the example is already sufficient, i think this sentence can be removed.

modules/concepts/pages/overrides.adoc
```diff
@@ -150,3 +150,73 @@ The `podOverrides` will be merged onto the following resources the operators dep
...
+
+1. All arguments listed in user `remove` are removed from operator
+2. All arguments matching any regex from user `removeRegex` are removed from operator
+3. All arguments from user `add` are added to operator
+
+You can check the resulting effective JVM arguments by looking at the ConfigMap containing the config for the roleGroup (although some tools read the JVM arguments from environmental variable).
```
maltesander, line 171:
```suggestion
You can check the resulting effective JVM arguments by looking at the ConfigMap containing the config for the roleGroup (although some tools read the JVM arguments from environmental variables).
```

modules/concepts/pages/overrides.adoc
```diff
@@ -150,3 +150,73 @@ The `podOverrides` will be merged onto the following resources the operators dep
 They will *not* be applied to:
 
 * Jobs, that are used to setup systems the product depends on e.g. create a database schema for Superset or Airflow.
+
+[#jvm-argument-overrides]
+== JVM argument overrides
+
+You can configure the JVM arguments used by JVM based tools.
+This is often times needed to e.g. configure a HTTP proxy or other network settings.
```
adwk67, line 158:
```suggestion
This is often needed to e.g. configure a HTTP proxy or other network settings.
```

modules/concepts/pages/overrides.adoc
```diff
@@ -150,3 +150,73 @@ The `podOverrides` will be merged onto the following resources the operators dep
...
+
+As with other overrides, the operator generates a set of JVM arguments that are needed to run the tool, you can specify additional arguments, which are merged on top of the ones the operator generated.
```
adwk67, line 160:
```suggestion
As with other overrides, the operator generates a set of JVM arguments that are needed to run the tool. You can specify additional arguments which are merged on top of the ones the operator generated.
```

modules/concepts/pages/overrides.adoc
adwk67, line 161:
```suggestion
As some JVM arguments are mutually exclusive (think of `-Xmx123m` and `-Xmx456m`), you also have the option to remove JVM arguments - either by specifying the exact argument or a regex.
```

modules/concepts/pages/overrides.adoc
```diff
@@ -150,3 +150,73 @@ The `podOverrides` will be merged onto the following resources the operators dep
...
+As some JVm arguments are mutually exclusive (think of `-Xmx123m` and `-Xmx456m`), you also have the option to remove JVM arguments - either by specifying the exact argument or a regex.
+
+The merging mechanism is applied <operator-generated> <- <role user specified> <- <rolegroup user specified> and works as following:
```
adwk67, line 163:
```suggestion
The merging mechanism is applied <operator-generated> <- <role user specified> <- <rolegroup user specified> and works as follows:
```

modules/concepts/pages/overrides.adoc
adwk67, line 171:
```suggestion
You can check the resulting effective JVM arguments by looking at the ConfigMap containing the config for the roleGroup (although some tools read the JVM arguments from an environmental variable).
```

adwk67 (changes requested):
Just minor things.

modules/concepts/pages/overrides.adoc
sbernauer: Makes sense. Please have a look at 8f8e78d

adwk67 (dismissed):
LGTM!

sbernauer:
Thanks for your quick review! The merge is blocked on the decision https://github.com/stackabletech/decisions/issues/37

modules/concepts/pages/overrides.adoc
```diff
@@ -150,3 +150,71 @@ The `podOverrides` will be merged onto the following resources the operators dep
...
+You can configure the JVM arguments used by JVM based tools.
+This is often needed to e.g. configure a HTTP proxy or other network settings.
+
+As with other overrides, the operator generates a set of JVM arguments that are needed to run the tool. You can specify additional arguments which are merged on top of the ones the operator generated.
+As some JVM arguments are mutually exclusive (think of `-Xmx123m` and `-Xmx456m`), you also have the option to remove JVM arguments - either by specifying the exact argument or a regex.
+
+The merging mechanism is applied <operator generated> <- <role user specified> <- <rolegroup user specified> and works as follows:
+
+1. All arguments listed in user specified `remove` are removed from operator generated
+2. All arguments matching any regex from user `removeRegex` are removed from operator generated
```
maltesander, line 166: We should mention that anchors `^{without_anchors}$` are set automatically?

sbernauer: Further down we mention
> # They need to match the entire string, not only a part!

I would like to avoid mentioning an internal detail (the regex crate has no function "does this string match this regex?". We only need the anchors because of this limitations.

WDYT of?

`All arguments matching any regex from user `removeRegex` are removed from operator generated. The regex needs to match the entire argument, not only a substring`

maltesander: Works for me yes!

sbernauer, line 166:
```suggestion
2. All arguments matching any regex from user removeRegex are removed from operator generated.
   The regex needs to match the entire argument, not only a substring
```

sbernauer: Thanks!

maltesander (approved):
LGTM!

PR #699, opened 2025-01-21
Add basic containerdebug logs

nightkr:


modules/concepts/pages/containerdebug.adoc
```diff
@@ -0,0 +1,23 @@
...
+environment, such as available disk interfaces and network interfaces.
+
+This logging is performed by our https://github.com/stackabletech/containerdebug[containerdebug] tool.
+
+NOTE: This tool is intended as a debugging aid, and the particular information or
+format should not be considered stable.
+
+== Interval
+
+This runs every minute. It can also be triggered manually, by running the
```
xeniape, line 13: Just two small nits: would prefer a more verbose description/naming/definition instead of "This" (also in the Output section below `This information`) because it's not completely clear what is meant here when reading, at least for me.

Everything else LGTM :smile:

xeniape (changes requested):

modules/concepts/pages/containerdebug.adoc
nightkr: Tried to clarify a bit in df262c93d4e08ec821e2fb8ab585d25dd2b72926, not sure how much it helps... >.<

xeniape: Perfect, thank you! :)

xeniape (approved):

lfrancke:
Do these docs need to be changed now that we moved the frequency?
https://github.com/stackabletech/containerdebug/pull/23

nightkr:
#710.

lfrancke:
Thank you

PR #701, opened 2025-01-22
Document higher temporary secret lifetime of Trino coordinators

sbernauer:


Maleware (approved):
Sweet! Thanks :) LGTM

PR #703, opened 2025-01-27
refactor: Restructure docs into partials and handle patch releases

Techassi:
- Move release notes into partials.
- Restructure headings to allow for patch releases.


NickLarsenNZ (approved):
LGTM

PR #707, opened 2025-02-14
fix: Refer supported k8s versions to release notes

adwk67:
For supported Kubernetes versions, refer to the release docs rather have an extra place that needs to be maintained.
See: 

- https://deploy-preview-707--stackable-docs.netlify.app/home/nightly/product-information/#_kubernetes
- https://deploy-preview-707--stackable-docs.netlify.app/home/nightly/kubernetes/#supported-production-distributions

This removes a point of redundancy, but does result in a link to all release info where one might expect to see information pertinent to the current release (which then has to be maintained separately to the release docs).


razvan (approved):
lgtm!

PR #712, opened 2025-02-25
docs: Add CRD versioning ADR

Techassi:
Part of https://github.com/stackabletech/issues/issues/642

### DIFF
--- a/modules/contributor/pages/adr/ADR000-crd-versioning.md
+++ b/modules/contributor/pages/adr/ADR000-crd-versioning.md
@@ -1,0 +1,563 @@
+The following is an initial (and still rough) draft of the CRD versioning ADR.
+It was written in Markdown using a HedgeDoc pad.
+This pad can be used for further collaboration if needed.
+The ADR will be converted into Asciidoc at the end (sad noises).
+
+# ADR000: CRD Versioning
+
+## Video notes
+
+{%youtube niamejFf5Zk %}
+
+- Alpha versions are considered "free for all"
+  - This means introducing breaking changes is possible
+  - No stability guarantees for users
+  - There is no need for automatic conversion using a webhook between alpha versions
+- Beta versions can also be breaking, but try to avoid them
+  - Instead, provide instructions on how to upgrade CRDs
+  - Automatic conversion from alpha to beta recommended
+- Stable versions, basically no changes, stable
+  - Automatic conversion from beta to stable recommended
+- Staying on alpha has its pros and cons:
+  - Pros:
+    - no extra code
+    - no need to bump
+    - good for bug fixes
+  - Cons:
+    - Not future-proof
+    - Can turn away enterprise users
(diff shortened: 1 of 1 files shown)


modules/contributor/pages/adr/ADR000-crd-versioning.md
```diff
@@ -0,0 +1,563 @@
...
+  foo: 42
+  bar: "hello, world" # Now required
+```
+
+### Optional struct field
+
+> [!TIP]
+> This approach can be used for fields which opt into optional (experimental) features but still providing the ability to customize using the field. Using `Option<T>` will push the responsibility of dealing with `None` (eg. using a default fallback value) to the operator.
+
+**TODO:** Can we detect the addition of an optional field and thus enable not needing to bump / introduce a new version? We *can* somehow reliable detect `Option`, but it is not guaranteed to be the `Option` (from the standard library) we expect.
```
sbernauer, line 111: I fear it's very complicated to determine this and I'm fine not automating this for now (and rely on human intelligence to determine if the version should bumped).
*If* we determine the breakingness however, I think we should be looking at the generated crd.yaml, not the Rust code. Maybe there are already some tools for that already ;)

modules/contributor/pages/adr/ADR000-crd-versioning.md
```diff
@@ -0,0 +1,563 @@
...
+spec:
+  foo: 42
+  bar: null # Explicitly setting the field to null
+```
+
+```yaml
+# ...
+spec:
+  foo: 42
+          # Not setting the field at all
```
sbernauer, line 142: nit (optional): I find it easier to read this way
```suggestion
  # Not setting the field at all
```

modules/contributor/pages/adr/ADR000-crd-versioning.md
```diff
@@ -0,0 +1,563 @@
...
+    #[versioned(added(since = "v1alpha2"))]
+    baz: usize,
+}
+```
+
+<!-- TODO: Add summary of the section clearly stating the way to handle this situation -->
+
+## Change: Renaming a field
+
+A renamed field can **always** be modelled as a non-breaking change which is also backwards compatible, meaning that a newer version of a field can be converted back to an older version.
```
sbernauer, line 196: > A renamed field can **always** be modelled as a non-breaking

From my perspective renaming a field is breaking. There are valid yamls before that are not valid after

sbernauer (commented):
LGTM in general, just some small remarks

modules/contributor/pages/adr/ADR000-crd-versioning.md
Techassi: Yes, this needs more clarification. I is only a non-breaking change if CRD conversion webhooks are in place. I've clarified this section in my current iteration.

PR #720, opened 2025-03-20
Add a internal page for every k8s distro

sbernauer:


modules/ROOT/pages/kubernetes/ibm-cloud.adoc
```diff
@@ -1,5 +1,7 @@
-= IBM cloud
+= IBM Cloud Kubernetes Service
 
-IBM Cloud uses a non-standard Kubelet state directory.
-Therefore, installing the secret-operator and listener-operator on IBM Cloud requires  special handling.
+https://www.ibm.com/products/kubernetes-service
+
+IBM Cloud Kubernetes Service uses a non-standard Kubelet state directory.
+Therefore, installing the secret-operator and listener-operator on IBM Cloud Kubernetes Service requires  special handling.
```
adwk67, line 6:
```suggestion
For this reason installing the secret-operator and listener-operator on IBM Cloud Kubernetes Service requires special handling.
```

adwk67 (changes requested):
The link to managed stackable seems to have changed: https://cloud.ionos.com/managed/managed-stackable

nightkr:
Hm, going through the extra page seems a bit overkill if we have nothing to add there anyway (though otoh, I can understand it as a stub for adding more info in the future...).

modules/ROOT/pages/kubernetes/ibm-cloud.adoc
sbernauer: Changed in 7b93ae4

sbernauer:
> (though otoh, I can understand it as a stub for adding more info in the future...).

Yeah, that was the idea. Ideally we get to a state where we have a getting started guide as we do for AKS or SKE

PR #731, opened 2025-04-17
docs: Add telemetry documentation page

Techassi:
Part of https://github.com/stackabletech/issues/issues/639

### DIFF
--- a/modules/concepts/nav.adoc
+++ b/modules/concepts/nav.adoc
@@ -22,7 +22,8 @@
 *** xref:operations/graceful_shutdown.adoc[]
 *** xref:operations/temporary_credentials_lifetime.adoc[]
 ** Observability
-*** xref:labels.adoc[]
-*** xref:logging.adoc[]
-*** xref:containerdebug.adoc[]
+*** xref:labels.adoc[Labels]
+*** xref:logging.adoc[Logging]
+*** xref:containerdebug.adoc[Container environment]
+*** xref:observability/telemetry.adoc[Telemetry]
 ** xref:container-images.adoc[]
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -22,10 +22,14 @@ For a more complete debugging experience, we recommend the OpenTelemetry signals
 ----
 telemetry:
   consoleLog:
-    enabled: true # true, false
-    level: null   # error, warning, info, debug, trace
-    format: null  # plain, json
+    enabled: true # <1>
+    level: null   # <2>
+    format: null  # <3>
 ----
+
+<1> Boolean: `true, false`
+<2> Enum: `error, warning, info, debug, trace`
+<3> Enum: `plain, json`
 
 == File logs
 
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -1,12 +1,10 @@
 = Telemetry signals in operators
 
-Since SDP 25.7.0, all Stackable operators emit multiple signals for telemetry data.
-Currently supported signals are:
+Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways:
 
-* Console logs in plain and JSON format
-* (Rolling) file logs in JSON format
-* OpenTelemetry logs exported via OTLP
-* OpenTelemetry traces exported via OTLP
+* logs (via console, file, and OTLP export)
+* metrics (not yet available, will most likely be an OTLP export)
+* traces (via OTLP export)
 
 Every signal can be toggled and customized using Helm values.
 All fields are grouped under the top-level `telemetry` section.
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -1,0 +1,32 @@
+= Telemetry signals in operators
+
+Since SDP 25.7.0, all Stackable operators emit multiple signals for telemetry data.
+Currently supported signals are:
+
+* Console logs in plain and JSON format
+* (Rolling) file logs in JSON format
+* OpenTelemetry logs exported via OTLP
+* OpenTelemetry traces exported via OTLP
+
+Every signal can be toggled and customized using Helm values.
+All fields are grouped under the top-level `telemetry` section.
+The following sections describe the available fields their default and supported values.
+
+== Console logs
+
+This signal prints log messages at the selected level and in the selected format to STDOUT.
+These logs are useful for quick debugging.
+For a more complete debugging experience, we recommend the OpenTelemetry signals.
+
+[source,yaml]
+----
+telemetry:
+  consoleLog:
+    enabled: true # true, false
+    level: null   # error, warning, info, debug, trace
+    format: null  # plain, json
+----
+
+== File logs
+
+== OpenTelemetry logs and traces
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -33,4 +33,22 @@ telemetry:
 
 == File logs
 
+This signal writes log messages at the selected level in JSON to (rolling) log file(s).
+These logs can be picked up by log aggregation systems, like Vector.
+
+[source,yaml]
+----
+telemetry:
+  fileLog:
+    enabled: false         # <1>
+    level: null            # <2>
+    rotationPeriod: hourly # <3>
+    maxFiles: 6            # <4>
+----
+
+<1> Boolean: `true, false`
+<2> Enum: `error, warning, info, debug, trace`
+<3> Enum: `never, daily, hourly, minutely`
+<4> Unsigned Integer
+
 == OpenTelemetry logs and traces
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -29,8 +29,8 @@ telemetry:
 ----
 
 <1> Boolean: `true, false`
-<2> Enum: `error, warning, info, debug, trace`
-<3> Enum: `plain, json`
+<2> Enum: `error, warning, info, debug, trace` - operator defaults to `info`
+<3> Enum: `plain, json` - operator defaults to `plain`
 
 == File logs
 
@@ -48,7 +48,7 @@ telemetry:
 ----
 
 <1> Boolean: `true, false`
-<2> Enum: `error, warning, info, debug, trace`
+<2> Enum: `error, warning, info, debug, trace` - operator defaults to `info`
 <3> Enum: `never, daily, hourly, minutely`
 <4> Unsigned Integer
 
@@ -71,5 +71,5 @@ telemetry:
 ----
 
 <1> Boolean: `true, false`
-<2> Enum: `error, warning, info, debug, trace`
+<2> Enum: `error, warning, info, debug, trace` - operator defaults to `info`
 <3> String: E.g. `my-collector:4317`
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -10,7 +10,8 @@ Currently supported signals are:
 
 Every signal can be toggled and customized using Helm values.
 All fields are grouped under the top-level `telemetry` section.
-The following sections describe the available fields their default and supported values.
+
+NOTE: The following sections describe the available fields their default and supported values.
 
 == Console logs
 
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -10,7 +10,19 @@ Since SDP 25.7.0, all Stackable operators emit telemetry data in the following w
 Every subscriber can be toggled and customized using Helm values.
 All fields are grouped under the top-level `telemetry` section.
 
-NOTE: The following sections describe the available fields their default and supported values.
+[NOTE]
+====
+The following sections describe the available fields as well as their default and supported values.
+If the Helm values explained below are not overridden, the following operator defaults apply:
+
+* Levels for all signals are set to `INFO`.
+* Console logs are enabled
+* File logs are disabled
+* File logs rotation period is set to `Never`
+* File logs max files is unset
+* OpenTelemetry logs and traces are disabled
+* OTLP endpoints are set to `http://localhost:4317`
+====
 
 == Console logs
 
@@ -28,8 +40,8 @@ telemetry:
 ----
 
 <1> Boolean: `true, false`
-<2> Enum: `error, warning, info, debug, trace` - operator defaults to `info`
-<3> Enum: `plain, json` - operator defaults to `plain`
+<2> Enum: `error, warning, info, debug, trace`
+<3> Enum: `plain, json`
 
 == File logs
 
@@ -47,7 +59,7 @@ telemetry:
 ----
 
 <1> Boolean: `true, false`
-<2> Enum: `error, warning, info, debug, trace` - operator defaults to `info`
+<2> Enum: `error, warning, info, debug, trace`
 <3> Enum: `never, daily, hourly, minutely`
 <4> Unsigned Integer
 
@@ -70,5 +82,5 @@ telemetry:
 ----
 
 <1> Boolean: `true, false`
-<2> Enum: `error, warning, info, debug, trace` - operator defaults to `info`
+<2> Enum: `error, warning, info, debug, trace`
 <3> String: E.g. `my-collector:4317`
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -2,10 +2,12 @@
 
 Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways:
 
-* logs (via console, file, and OTLP export)
-* traces (via OTLP export)
+* Console logs in plain and JSON format
+* File logs in JSON format (with the option of Rolling files)
+* OpenTelemetry logs exported via OTLP
+* OpenTelemetry traces exported via OTLP
 
-Every signal can be toggled and customized using Helm values.
+Every subscriber can be toggled and customized using Helm values.
 All fields are grouped under the top-level `telemetry` section.
 
 NOTE: The following sections describe the available fields their default and supported values.
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -46,7 +46,7 @@ telemetry:
 == File logs
 
 This signal writes log messages at the selected level in JSON to (rolling) log file(s).
-These logs can be picked up by log aggregation systems, like Vector.
+These logs can be picked up by a log aggregation system, like Vector.
 
 [source,yaml]
 ----
--- a/modules/concepts/pages/observability/telemetry.adoc
+++ b/modules/concepts/pages/observability/telemetry.adoc
@@ -52,3 +52,23 @@ telemetry:
 <4> Unsigned Integer
 
 == OpenTelemetry logs and traces
+
+These two signal export OpenTelemetry logs and traces to OTLP enabled collectors.
+These signals can be visualized using tools, like Grafana, Loki, and Jaeger.
+
+[source,yaml]
+----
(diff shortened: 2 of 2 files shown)


modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,74 @@
+= Telemetry signals in operators
+
+Since SDP 25.7.0, all Stackable operators emit multiple signals for telemetry data.
```
NickLarsenNZ, line 3: We should maybe rephrase this. The signals are:

- logs (with 3 options: console, file, otlp export)
- metrics (not yet available, but might be an openmetrics server and/or otlp export)
- traces

```suggestion
Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways.
```

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,74 @@
...
+Currently supported signals are:
+
+* Console logs in plain and JSON format
+* (Rolling) file logs in JSON format
+* OpenTelemetry logs exported via OTLP
+* OpenTelemetry traces exported via OTLP
```
NickLarsenNZ, line 9: I feel like Rolling is a feature that can be enabled

```suggestion
* Console logs in plain and JSON format
* File logs in JSON format (with the option of Rolling files)
* OpenTelemetry logs exported via OTLP
* OpenTelemetry traces exported via OTLP
```

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+Every signal can be toggled and customized using Helm values.
+All fields are grouped under the top-level `telemetry` section.
+The following sections describe the available fields their default and supported values.
```
NickLarsenNZ, line 13: Maybe a new paragraph to keep the line shorter since it contains import info at the end.

```suggestion
Every signal can be toggled and customized using Helm values.
All fields are grouped under the top-level `telemetry` section.

The following sections describe the available fields their default and supported values.
```

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: I'll redo this comment as it needs to include the next line

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,74 @@
+= Telemetry signals in operators
+
+Since SDP 25.7.0, all Stackable operators emit multiple signals for telemetry data.
+Currently supported signals are:
```
NickLarsenNZ, line 4: We should maybe rephrase this. The signals are:

- logs (with 3 options: console, file, otlp export)
- metrics (not yet available, but might be an openmetrics server and/or otlp export)
- traces

```suggestion
Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways:
```

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: See: https://github.com/stackabletech/documentation/pull/731/files#r2058284411

modules/concepts/pages/observability/telemetry.adoc
Techassi: Done in e230821cb287dae533b8ae00048da9b4225a227b and f805aceed4e0db104b01dfc7d2d99685356b6f92.

modules/concepts/pages/observability/telemetry.adoc
Techassi: Done in 732e0625ea5d5bea1c5192d460353c3aa6b90670.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,73 @@
+= Telemetry signals in operators
+
+Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways:
+
+* logs (via console, file, and OTLP export)
+* metrics (not yet available, will most likely be an OTLP export)
+* traces (via OTLP export)
```
NickLarsenNZ, line 7: I think it was fine how you had it. I was just explaining signals in the comment.

But this is also fine, but we should remove metrics (we can add an admonition to mention metrics would come later).

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,73 @@
...
+telemetry:
+  consoleLog:
+    enabled: true # <1>
+    level: null   # <2>
+    format: null  # <3>
+----
+
+<1> Boolean: `true, false`
+<2> Enum: `error, warning, info, debug, trace` - operator defaults to `info`
+<3> Enum: `plain, json` - operator defaults to `plain`
```
NickLarsenNZ, line 31: For each of these we should list the operator defaults.

We could just put it at the top and say that "this is what happens if the subscriber is not configured".

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: For this one, my comment wasn't the verbatim suggestion, just pointing out the difference between the subscribers and signals.

modules/concepts/pages/observability/telemetry.adoc
Techassi: Done in e3fc6e824d4160984bbf312e0a5d3d4bfdf9489f.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways:
+
+* logs (via console, file, and OTLP export)
+* traces (via OTLP export)
+
+Every signal can be toggled and customized using Helm values.
+All fields are grouped under the top-level `telemetry` section.
+
+NOTE: The following sections describe the available fields their default and supported values.
```
NickLarsenNZ, line 11: Sorry, I think my original comment wasn't clear enough.
I think how you originally had it was good, just needed to not call them "subscribers" instead of "signals".

Here are two options to illustrate what I mean:

## Original list

_with my amendment for the File logs_:

```suggestion
Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways:

* Console logs in plain and JSON format
* File logs in JSON format (with the option of Rolling files)
* OpenTelemetry logs exported via OTLP
* OpenTelemetry traces exported via OTLP

Every subscriber can be toggled and customized using Helm values.
All fields are grouped under the top-level `telemetry` section.

NOTE: The following sections describe the available fields their default and supported values.
```

## Original first line(s)

_but then listing signals, and not subscribers_:

```suggestion
Since SDP 25.7.0, all Stackable operators emit multiple signals for telemetry data.
Currently supported signals are:

* logs (via console, file, and OTLP export)
* traces (via OTLP export)

Every subscriber can be toggled and customized using Helm values.
All fields are grouped under the top-level `telemetry` section.

NOTE: The following sections describe the available fields their default and supported values.
```

NickLarsenNZ (changes requested):
Clarified what I meant with the signals vs subscribers

modules/concepts/pages/observability/telemetry.adoc
Techassi: Adjusted in 62bd5635.

modules/concepts/pages/observability/telemetry.adoc
Techassi: Ah, gotcha. I think it should be fine now since 4da9817.

NickLarsenNZ (approved):
LGTM

modules/concepts/nav.adoc
```diff
@@ -22,7 +22,8 @@
 *** xref:operations/graceful_shutdown.adoc[]
 *** xref:operations/temporary_credentials_lifetime.adoc[]
 ** Observability
-*** xref:labels.adoc[]
-*** xref:logging.adoc[]
-*** xref:containerdebug.adoc[]
+*** xref:labels.adoc[Labels]
+*** xref:logging.adoc[Logging]
+*** xref:containerdebug.adoc[Container environment]
```
nightkr, line 27: These labels are redundant, no? They should be pulled in automatically since they're the top headings of their respective pages.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,86 @@
+= Telemetry signals in operators
```
nightkr, line 1: I'd put operator logging as a subheader in `logging.adoc` instead, I think? Yes, "telemetry" is technically slightly more general, but "logging" is the term people know and love already (and the current split of "logging as in products, telemetry as in operators" makes even less sense IMO).

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+Since SDP 25.7.0, all Stackable operators emit telemetry data in the following ways:
```
nightkr, line 3: Same here, I think it's clearer to reword this focusing on "logging" rather than the pretty opaque "telemetry data". Once we do offer non-log telemetry (metrics), we can have an "oh, and we also export metrics via OTLP and Prom" note at the end (or just treat metrics as entirely separate).

nightkr: (And yes, I'd argue that "traces" are just logs with a higher standard of correlation information, not a distinct concept.)

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+* Console logs in plain and JSON format
+* File logs in JSON format (with the option of Rolling files)
+* OpenTelemetry logs exported via OTLP
+* OpenTelemetry traces exported via OTLP
+
+Every subscriber can be toggled and customized using Helm values.
```
nightkr, line 10: s/subscriber/logging endpoint/g? There's still probably a better word hiding somewhere in there though.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,86 @@
...
+telemetry:
+  consoleLog:
+    enabled: true # <1>
+    level: null   # <2>
+    format: null  # <3>
+----
+
+<1> Boolean: `true, false`
+<2> Enum: `error, warning, info, debug, trace`
+<3> Enum: `plain, json`
```
nightkr, line 44: It feels very disjointed to have to jump back and forth between these, maybe just add the options inline? (`enabled: true # true/false`)

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+== File logs
+
+This signal writes log messages at the selected level in JSON to (rolling) log file(s).
+These logs can be picked up by log aggregation systems, like Vector.
```
nightkr, line 49:
```suggestion
These logs can be picked up by a log aggregation systems, like Vector.
```

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -0,0 +1,86 @@
...
+
+[source,yaml]
+----
+telemetry:
+  fileLog:
+    enabled: false         # <1>
+    level: null            # <2>
```
nightkr, line 56: M'yeah, I get why these are treated as separate ("just jump to the platform you use, and go!"), but it also feels a bit strange that we treat these as entirely separate. I don't know, I don't have a good answer here.

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: > Same here, I think it's clearer to reword this focusing on "logging" rather than the pretty opaque "telemetry data"

I disagree. I think we want to be promoting "observability" of the platform and move on from "Logging" being the central term. It is merely one of the telemetry signals.

> And yes, I'd argue that "traces" are just logs with a higher standard of correlation information

No, they are another telemetry signal. Otherwise they'd just be called _logs_.

> Once we do offer non-log telemetry (metrics), we can have an "oh, and we also export metrics via OTLP and Prom" note at the end (or just treat metrics as entirely separate).

Even though we don't yet have operators emitting metrics, I think we can still begin advertising "observability" capabilities. Once metrics are available, they can be added into this documentation.

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: @nightkr I edited your suggetion (removing the superfluous _s_ from _systems_)

modules/concepts/pages/observability/telemetry.adoc
nightkr: > No, they are another telemetry signal. Otherwise they'd just be called logs.

Okay, what's the conceptual difference? When, in your view, would traditional flat logs still be preferable?

modules/concepts/pages/observability/telemetry.adoc
Techassi: Yes, I'd be in favour of calling them _exporters_.

nightkr: Also, "log to file" isn't really a useful option in its own right IMO (we're only writing to a _container_ file, after all), that's more of an implementation detail for shipping to something like Vector.

modules/concepts/nav.adoc
Techassi: Currently, they are.

I added them because I moved the files into the `observability` subfolder, but that broke links in other repos (operators for example). So I reverted that change, but kept the labels for the future when we eventually move them again (which I think we should).

modules/concepts/pages/observability/telemetry.adoc
Techassi: > "logging" is the term people know and love already

I would like to avoid bundling all these signal into _logging_. It is more than that already, and will differ even more once we add support for metrics (or even [profiling](https://opentelemetry.io/blog/2024/profiling/)).

> logging as in products, telemetry as in operators

There will be a transition period. Eventually, we also want to support / enable telemetry data for products. This will then form a complete picture, which can also be reflected in the documentation accordingly.

But for now, there are differences between operators and products which are worth to be pointed out.

modules/concepts/pages/observability/telemetry.adoc
Techassi: I actually had that at one point. I've converted them to _callouts_ to be a little more consistent with the rest of out documentation.

But I don't have strong feelings for either option.

modules/concepts/pages/observability/telemetry.adoc
nightkr: > I would like to avoid bundling all these signal into logging.

Tracing *is* (an extended form of) logging.

Metrics and profiling are not. They (would) deserve their own sections.

Techassi: Tracing _can_ be considered an extended form of logging, but the OpenTelemetry project specifies them as two different [signals](https://opentelemetry.io/docs/concepts/signals/). As such, I will follow the common terminology and will not mix both together to avoid ambiguity.

modules/concepts/pages/observability/telemetry.adoc
Techassi: These log files can also be picked up by other log scraping solutions users might want to implement. For _us_, yes, there are mostly only useful to feed Vector.

modules/concepts/pages/observability/telemetry.adoc
Techassi: Accepting this suggestion after the edit.

modules/concepts/pages/observability/telemetry.adoc
nightkr: No, this is a vital distinction: the documentation is defining our *supported interfaces*. Do *we* support scraping the raw JSON files? Is that a stable format that we're committing to? How do we suggest that people hook in to scrape it?

modules/concepts/pages/observability/telemetry.adoc
nightkr: The different logging appenders, from a documentation perspective.

Currently they're documented as "We provide these four appenders. Each appender supports [these options].". I'm *wondering* if it might be clearer/more maintainable to write it as "We support these appenders, which can all be enabled/filtered [like this]. Each appender msut be connected to a target, [like this].". (That is, focus on the commonalities, *then* branch off into the particulars of each appender.)

modules/concepts/pages/observability/telemetry.adoc
nightkr: Logs can still be structured (https://github.com/slog-rs/slog, https://datalust.co/seq, etc). Traces are "just" a convention for including more context. But it's not a fundamental change.

> For most things, I would consider traditional logs rather cumbersome to work with and as such are not preferable over more structured data like traces (and metrics). Traditional logs are mostly only useful for a quick debugging session.

Exactly. Because it's just "Logs 2.0", not a distinct new concept.

NickLarsenNZ: I'm not sure what your point is here, but it seems like "I can make one data structure look like something else to imply they are the same in one case", and then coining distracting terms like "Logs 2.0" which isn't a thing.

Tracing is a conceptually different way of observing a system, no matter how it is serialized or transported. We should use the common terminology which isn't new anymore. Logging is one part.

IMO, structured logs are largely eliminated by tracing events.
I personally don't see value in logs that don't correlate to a span - especially when troubleshooting. But I'm also aware that people still cling to logs, hence why they are kept as an independent signal (the same content appearing in trace events, which the "structured" fields becoming span/resource attributes).

Techassi: > Logs can still be structured

Sure, I'm not arguing against that, but that still doesn't mean that structured logs are the same thing as traces. Calling them "Logs 2.0" undercuts the value they provide.

That's why I won't bundle (traditional) logs and traces together as already stated in https://github.com/stackabletech/documentation/pull/731#discussion_r2063747531.

modules/concepts/pages/observability/telemetry.adoc
Techassi: It should be noted, that it is currently not possible to use any of the (rolling) file log functionality via Helm. The required env vars are just not set by the Deployment template. We corrected this, by introducing this control knob.

> Is that a stable format that we're committing to?

We don't _have_ to make any promises regarding the log format in my opinion. This can also be pointed out in the documentation. _If_ we want to make a commitment, we can discuss the details of this in a followup issue/PR (needs research).

> How do we suggest that people hook in to scrape it?

I would say this is outside of the current PR. Also, I don't see the need to go into this too deep, as we want to promote the OpenTelemetry appoach anyway. But this is also something which can be tackled in a followup PR (needs refinement and planning).

modules/concepts/pages/observability/telemetry.adoc
Techassi: > The different logging appenders

We are not only talking about logging appenders. There is a console log subscriber, a file log appender, as well as log and traces exporters. That's why @NickLarsenNZ suggested to call these consumers _exporters_ to have a unified name without exposing the users to underlying crate details, see https://github.com/stackabletech/documentation/pull/731#discussion_r2063096186.

> I'm wondering if it might be clearer/more maintainable to write it as "We support these appenders, which can all be enabled/filtered [like this] [...] That is, focus on the commonalities

Okay, I can see your point. We _can_ do that, but this will result in "jumping back and forth" for the reader, which you pointed out as a negative point in https://github.com/stackabletech/documentation/pull/731#discussion_r2059921806.

So let me know how you weight the pros and cons here.

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: _Speaking of stability in general, I think we gain some stability/predictability by using field names that align with [OpenTelemetry Semantic Conventions]. This will help them correlate with logs/traces/metrics emitted by external tools that also move to align to [OpenTelemetry Semantic Conventions]._

NickLarsenNZ (approved):
LGTM

lfrancke:
Can you please add a link to the docs?

Techassi:
Sure, here it is: <https://docs.stackable.tech/home/nightly/concepts/observability/telemetry/>.

lfrancke:
And a release note snippet please

PR #740, opened 2025-05-07
Add note on telemetry and call-home functionality

sbernauer:
There is some folder restructuring in the first commit, I suggest to review them individually


modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -1,4 +1,10 @@
-= Telemetry signals in operators
+= Operational telemetry signals in operators
```
Techassi, line 1: note: I think I like the original title better.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -1,4 +1,10 @@
...
+
+[NOTE]
+====
+Telemetry data is useful for your own observability and monitoring systems.
+There is no "call-home" functionality, data is not exported out of the container unless explicitly configured.
+====
```
Techassi, line 7: suggestion: Add a full stop.

```suggestion
[NOTE]
====
Telemetry data is useful for your own observability and monitoring systems.
There is no "call-home" functionality.
Data is not exported out of the container unless explicitly configured.
====
```

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -1,4 +1,11 @@
-= Telemetry signals in operators
+= Operational telemetry signals in operators
```
sbernauer, line 1:
```suggestion
= Telemetry signals in operators
```

modules/concepts/pages/observability/telemetry.adoc
sbernauer: reverted

PR #741, opened 2025-05-21
docs(code-style-guide): Add section about unwraps

Techassi:
This PR adds two sections about `unwrap` and why it is not recommended to be used.

### DIFF
--- a/modules/contributor/pages/code-style-guide.adoc
+++ b/modules/contributor/pages/code-style-guide.adoc
@@ -531,6 +531,41 @@ enum Error {
 . `unable to read config file from ...` to indicate that the file could not be loaded (for example because the file doesn't exist).
 . `unable to parse value ...` to indicate that parsing a user provided value failed (for example because it didn't conform to the expected syntax).
 
+=== Using `unwrap`
+
+Generally, it is not recommended to use `unwrap` (or any other method which consumes the error) in any fallible code path.
+Instead, proper error handling like above should be used.
+There are however cases, where it is fine to use `unwrap` or friends.
+
+One such an example is when compiling regular expressions inside const/static environments.
+For such cases code must use `expect` instead of `unwrap` to provide additional context why a particular piece of code should never fail.
+
+// Do we want to mention that this is enforced via clippy and that we actually enable that lint in our repos?
+
+[TIP.code-rule,caption=Examples of correct code for this rule]
+====
+
+[source,rust]
+----
+static VERSION_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r".*").expect("failed to compile regex")
+});
+----
+
+====
+
+[WARNING.code-rule,caption=Examples of incorrect code for this rule]
+====
+
+[source,rust]
+----
+static VERSION_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r".*").unwrap()
+});
+----
+
+====
+
 == String formatting
 
 === Named versus unnamed format string identifiers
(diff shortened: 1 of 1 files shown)


modules/contributor/pages/code-style-guide.adoc
```diff
@@ -737,3 +772,27 @@ mod test {
 ----
 
 ====
+
+=== Using `unwrap`
+
+The usage of `unwrap` in unit tests is not recommended for the same reasons as mentioned above, but allowed.
```
NickLarsenNZ, line 778: I think we should not allow unwrap in tests for exactly the same reason. It could come later, but we should reword the style guide to not allow in in the first place.

```suggestion
The usage of `unwrap` in unit tests is also not allowed for the same reasons as mentioned above.
```

If we get to the point that running code no longer uses unwrap, ie `#![deny(clippy::unwrap_used)]`, we can permit it for tests until they are sorted:

```rust
#[cfg(test)]
#[allow(clippy::unwrap_used)]
mod test {
  // ...
}
```

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -531,6 +531,41 @@ enum Error {
 . `unable to read config file from ...` to indicate that the file could not be loaded (for example because the file doesn't exist).
 . `unable to parse value ...` to indicate that parsing a user provided value failed (for example because it didn't conform to the expected syntax).
 
+=== Using `unwrap`
+
+Generally, it is not recommended to use `unwrap` (or any other method which consumes the error) in any fallible code path.
+Instead, proper error handling like above should be used.
+There are however cases, where it is fine to use `unwrap` or friends.
```
NickLarsenNZ, line 538: _incomplete suggestion..._

```suggestion
The `unwrap` function must not be used in any fallible code paths.
Instead, proper error handling like above should be used.

There are however cases, where it is fine to use `unwrap` or friends.
```

I want to mention something line `unwrap_or` and friends are fine, since they will not panic.

The main things I want to point out here are:
- We should ban `unwrap` entirely` (eventually denying via clippy).
- Error handling done like above, unless there is valid reason to use `expect` as described below.

NickLarsenNZ (commented):
Incomplete review. Happy to help on a screenshare.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -531,6 +531,41 @@ enum Error {
...
+
+One such an example is when compiling regular expressions inside const/static environments.
+For such cases code must use `expect` instead of `unwrap` to provide additional context why a particular piece of code should never fail.
```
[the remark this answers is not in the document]
Techassi: > Technically even static environments could be unsafe (LazyLock using sometime at runtime), but static strings would be fine.

That's true. But most of our current static usages don't rely on runtime values. We can however re-visit this in the future and refine the rule.

I applied your second suggestion.

modules/contributor/pages/code-style-guide.adoc
Techassi: I've updated your suggestion and applied it.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -531,6 +531,41 @@ enum Error {
...
+
+// Do we want to mention that this is enforced via clippy and that we actually enable that lint in our repos?
```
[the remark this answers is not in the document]
Techassi: > We should first do a sweep of our code, as there are plenty of unwraps that need either graceful error handling or converted to expect.

I did a quick sweep locally using `rg` across all Rust repositories I have cloned:

```
914 matches
904 matched lines
202 files contained matches
```

This is pretty substantial.

> Then we can decide when to completely ban unwrap.

I would definitely go for this option. Using `unwrap` is a pretty bad smell in my opinion.

modules/contributor/pages/code-style-guide.adoc
```diff
@@ -737,3 +776,35 @@ mod test {
...
+The usage of `unwrap` in unit tests is also not allowed for the same reasons as mentioned above.
+
+[TIP.code-rule,caption=Examples of correct code for this rule]
+====
+
+[source,rust]
+----
+#[test]
+fn deserialize() {
+    let input: String = serde_yaml::from_str("my string").expect("input string must deserialize");
```
NickLarsenNZ, line 791:
```suggestion
    let input: String = serde_yaml::from_str("my string").expect("constant input string must deserialize");
```

NickLarsenNZ (approved):
Lgtm

PR #745, opened 2025-06-05
Add a note about using nix-docker-builder when on macOS/Darwin

nightkr:
This only covers configuring the remote builder, I have a separate PR in progress for operator-templating to actually use it.


Maleware (approved):
apart from the open question LGTM!

PR #762, opened 2025-07-24
refactor: Improve branch/release scripts

Techassi:
This PR improves a bunch of things:

- The release branches in the documentation repository now finally uses `release-YY.M.X` instead of `release/YY.M.X` to be inline with all other release branches.
- Converts all variable names to uppercase and improves the naming of some of them.
- Reports the commit of `HEAD` and `origin/main` if they are not in sync
- Print out the branch the PR should be raised against

### DIFF
--- a/scripts/make-release-branch.sh
+++ b/scripts/make-release-branch.sh
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# This script creates a new 'release/{major}.{minor}'  branch for the documetation,
-# off of the 'main' branch. 
+# This script creates a new 'release-{major}.{minor}'  branch for the documetation,
+# off of the 'main' branch.
 #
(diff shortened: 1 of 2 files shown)


NickLarsenNZ (approved):
LGTM.

Does it also need changes in the antora yamls?

---

Also:
- Will need to checkout branches, and push them with the new name
- Will need to update branch protection rules (if any)
- Update any templates that mention using the `/` notation

Techassi:
> Does it also need changes in the antora yamls?

No, the new release branch is inserted as was before. It only uses a hyphen instead of a slash.

> Will need to checkout branches, and push them with the new name

No, the old branches can be kept as is. All releases up to 25.3 can continue to use the branches with slashes. The new 25.7 release will instead point to a branch with a hyphen.

> Update any templates that mention using the / notation

Yes, that's true. But there shouldn't be many mentions afair.

PR #764, opened 2025-07-25
chore: Remove OpenSearch from 25.7

Techassi:
OpenSearch being there led to errors and we don't know why. This is the only solution which made the build work again.


NickLarsenNZ:
Mentioned it here: https://github.com/stackabletech/issues/issues/723#issuecomment-3116857743

Issue #756, opened 2025-07-16
Troubleshooting guide for stuck Pods through PDB restrictions


Maleware:
We currently have an Issue https://github.com/stackabletech/issues/issues/745 which needs a troubleshooting guide for customers.

## Acceptance Criteria
- [x] Documentation contains a user-oriented step-by-step guide to redeem the Problem
- [x] Documentation gives a suggestion on how to prevent the issue

Maleware:
backport https://github.com/stackabletech/documentation/pull/770

PR #769, opened 2025-08-04
Add documentation about PDBs

Maleware:
resolves https://github.com/stackabletech/documentation/issues/756

### DIFF
--- a/modules/concepts/pages/operations/pod_disruptions.adoc
+++ b/modules/concepts/pages/operations/pod_disruptions.adoc
@@ -1,5 +1,6 @@
 = Allowed Pod disruptions
 :k8s-pdb: https://kubernetes.io/docs/tasks/run-application/configure-pdb/
+:commons-operator: xref:commons-operator:index.adoc
 :description: Configure PodDisruptionBudgets (PDBs) to minimize planned downtime for Stackable products. Default values are based on fault tolerance and can be customized.
 
 Any downtime of our products is generally considered to be bad.
@@ -117,3 +118,35 @@ This PDB allows only one Pod out of all the Namenodes and Journalnodes to be dow
 
 == Details
 Have a look at the xref:contributor:adr/ADR030-allowed-pod-disruptions.adoc[ADR on Allowed Pod disruptions] for the implementation details.
+
+== Known issues with PDBs and e.g. certificate rotations
+PDBs together with certificate rotations can be problematic in case e.g. {commons-operator}[commons-operator] was unavailable until certificates expire. Assume a product like xref:zookeeper:index.adoc[Apache ZooKeeper] which needs a quorum and a PDB of `1`. If commons-operator comes back, it will rotate all expired certificates and thus `evict` pods. PDBs count exclusively for `eviction` and as the certificates are expired this pod will be stuck in `CrashLoopBackOff` forever unable to connect to the quorum, blocking PDB for the other pods. Other pods are unable to restart and rotate certificates.
+
+NOTE: We encountered this problem only with the specific outlined case above and only under this circumstances.
+
+=== Workaround
+If you encounter this only manually deleting those pods can help out of this situation as it is no `eviction`.
+
+==== k9s
+If you are using `k9s` you can start it in your terminal
+[source, bash]
+----
+k9s
+----
+and type `0` to view all namespaces and then type e.g. `/zookeeper` and hit enter. Go with up and down to your pod and press `CTL + D` and confirm to delete the pod. Repeat with all other instances of the stuck product.
+
+==== kubectl
+List your pods with
+[source, bash]
+----
+kubectl get pods -A
+----
+and copy the name of the pod you want to delete. Type
+[source, bash]
+----
+kubectl delete pod zookeeper-server-default-0
+----
+to delete the instance with the name `zookeeper-server-default-0`. Repeat it for all instances of your product.
+
+=== Preventing this Situation entirely
+Only disabling PDBs can guarantee to not run into this scenario.


modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -117,3 +118,35 @@ This PDB allows only one Pod out of all the Namenodes and Journalnodes to be dow
 
 == Details
 Have a look at the xref:contributor:adr/ADR030-allowed-pod-disruptions.adoc[ADR on Allowed Pod disruptions] for the implementation details.
+
+== Known issues with PDBs and e.g. certificate rotations
```
sbernauer, line 122: I *think* this is the only known issue, but I might be wrong
```suggestion
== Known issue with PDBs and certificate rotations
```

modules/concepts/pages/operations/pod_disruptions.adoc
sbernauer, line 1: Can we please rename the PR to something like `Add known issue on PDBs and certificate rotation`

modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -117,3 +118,35 @@ This PDB allows only one Pod out of all the Namenodes and Journalnodes to be dow
...
+PDBs together with certificate rotations can be problematic in case e.g. {commons-operator}[commons-operator] was unavailable until certificates expire. Assume a product like xref:zookeeper:index.adoc[Apache ZooKeeper] which needs a quorum and a PDB of `1`. If commons-operator comes back, it will rotate all expired certificates and thus `evict` pods. PDBs count exclusively for `eviction` and as the certificates are expired this pod will be stuck in `CrashLoopBackOff` forever unable to connect to the quorum, blocking PDB for the other pods. Other pods are unable to restart and rotate certificates.
```
sbernauer, line 123: note: For docs we normally have a newline after each sentence
```suggestion
PDBs together with certificate rotations can be problematic in case e.g. {commons-operator}[commons-operator] was unavailable to restart the Pods before the certificate expire.
commons-operator uses the `evict` API in Kubernetes, which respects the PDB.
If a Pod is evicted and a PDB would be violated, the Pod is *not* restarted.
Assume a product like xref:zookeeper:index.adoc[Apache ZooKeeper] which needs to form a quorum to function and the PDB only allows a single Pod to be unavailable.
As soon as enough certificates of the ZookeeperCluster have expired, all Pods will crash-loop, as they encounter expired certificates.
As only the container crash-loops (not the entire Pod), no new certificate is issued.
As soon as commons-operator comes online again it tries to `evict` a Zookeeper Pod.
However, this is prohibited, as the PDB would be violated.
```

modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -117,3 +118,35 @@ This PDB allows only one Pod out of all the Namenodes and Journalnodes to be dow
...
+List your pods with
+[source, bash]
+----
+kubectl get pods -A
+----
+and copy the name of the pod you want to delete. Type
+[source, bash]
+----
+kubectl delete pod zookeeper-server-default-0
+----
```
sbernauer, line 148: I think users running into this are advanced enough to delete Pods.
I would simply mention this command and call it a day
```bash
kubectl delete pod -l app.kubernetes.io/name=zookeeper,app.kubernetes.io/instance=simple-zk
pod "simple-zk-server-default-0" deleted
pod "simple-zk-server-default-1" deleted
pod "simple-zk-server-default-2" deleted
```

modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -117,3 +118,35 @@ This PDB allows only one Pod out of all the Namenodes and Journalnodes to be dow
...
 == Details
 Have a look at the xref:contributor:adr/ADR030-allowed-pod-disruptions.adoc[ADR on Allowed Pod disruptions] for the implementation details.
+
+== Known issues with PDBs and e.g. certificate rotations
+PDBs together with certificate rotations can be problematic in case e.g. {commons-operator}[commons-operator] was unavailable until certificates expire. Assume a product like xref:zookeeper:index.adoc[Apache ZooKeeper] which needs a quorum and a PDB of `1`. If commons-operator comes back, it will rotate all expired certificates and thus `evict` pods. PDBs count exclusively for `eviction` and as the certificates are expired this pod will be stuck in `CrashLoopBackOff` forever unable to connect to the quorum, blocking PDB for the other pods. Other pods are unable to restart and rotate certificates.
+
+NOTE: We encountered this problem only with the specific outlined case above and only under this circumstances.
+
+=== Workaround
+If you encounter this only manually deleting those pods can help out of this situation as it is no `eviction`.
```
sbernauer, line 128:
```suggestion
If you encounter this only manually deleting those pods can help out of this situation.
A Pod deletion (other than evictions) does *not* respect PDBs, so the Pods can be restarted anyway.
All restarted Pods will get a new certificate, the stacklet should turn healthy again.
```

modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -117,3 +118,35 @@ This PDB allows only one Pod out of all the Namenodes and Journalnodes to be dow
...
+kubectl get pods -A
+----
+and copy the name of the pod you want to delete. Type
+[source, bash]
+----
+kubectl delete pod zookeeper-server-default-0
+----
+to delete the instance with the name `zookeeper-server-default-0`. Repeat it for all instances of your product.
+
+=== Preventing this Situation entirely
```
sbernauer, line 151:
```suggestion
=== Preventing this situation
```

modules/concepts/pages/operations/pod_disruptions.adoc
```diff
@@ -117,3 +118,35 @@ This PDB allows only one Pod out of all the Namenodes and Journalnodes to be dow
...
+Only disabling PDBs can guarantee to not run into this scenario.
```
sbernauer, line 152:
```suggestion
The best measure is to make sure that commons-operator is always running, so that it can restart the Pods before the certificates expire.

A hacky way to prevent this situation could be to disable PDBs for the specific stacklet.
But this also has the downside, that you are now missing the benefits of the PDB.
```

sbernauer (changes requested):

modules/concepts/pages/operations/pod_disruptions.adoc
Maleware: done

modules/concepts/pages/operations/pod_disruptions.adoc
Maleware: https://github.com/stackabletech/documentation/pull/769/commits/b81d587e16d38e5c258678aa1ff393b6a3dad256

PR #770, opened 2025-08-05
Add known issue on PDBs and certificate rotation (backport)

Maleware:
Backports docu on PDBs from https://github.com/stackabletech/documentation/pull/769


sbernauer (approved):
I think I would prefer the PR title to be `Add known issue on PDBs and certificate rotation (backport)` but I leave it to you.
I don't think release 25.7 needs to be mentioned, as this commit will only be on the 25.7 release branch

PR #780, opened 2025-09-01
Add guide on how to debug network traffic

sbernauer:


modules/guides/pages/debug-network-traffic.adoc
```diff
@@ -0,0 +1,154 @@
...
+            args:
+              - -c
+              # If the dump grows to big, you can use regular tcpdump filters here
+              # to filter the captured traffic
+              - tcpdump -i any -w /tmp/tcpdump.pcap
+----
+
+=== Attach without restart
+
+You can also use something like `kubectl debug trino-coordinator-default-0 -it --image=nicolaka/netshoot -c tcpdump` to use a debug container and attach to a Pod without restart.
```
NickLarsenNZ, line 48: FYI, I didn't try this, but will separately next time I need it.

modules/guides/pages/debug-network-traffic.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+== TLS decryption usage
+
+Let's make things a bit more interesting using a real-world example.
+As Superset is behaving weird, we want to debug the network traffic from Superset to Trino, which is using HTTPS.
```
NickLarsenNZ, line 53:
```suggestion
Let's assume Superset is behaving weird and we want to debug the network traffic from Superset to Trino, which is using HTTPS.
```

NickLarsenNZ (dismissed):
I didn't carefully check all text, but this worked for me.

Thanks so much!

modules/guides/pages/debug-network-traffic.adoc
sbernauer: If you feel fancy, the TLS sniffer agent *should* also support attaching to a running process. And I *think* containers share a process namespace, but not sure

PR #825, opened 2026-02-25
Extend list for products supporting OPA

maltesander:
Added missing products:

- Airflow
- HBase
- HDFS
- NiFi
- Superset


xeniape (changes requested):
The ui submodule seems to be looking at an outdated version
Edit: I think you had to run `git submodule update --remote`

xeniape (approved):
LGTM

PR #834, opened 2026-04-15
fix: Correct the DruidCluster bucket reference config in S3 docs

xeniape:
Fix the bucket reference configuration in example


razvan (approved):
lgtm

PR #835, opened 2026-04-16
Adding documentation on otel with vector

Maleware:
This PR adds documentation about how to connect otel logs emitted by operators to a vector sink matching the product logs.


modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
 <1> Boolean: `true`, `false`
 <2> String: `error`, `warn`, `info`, `debug`, `trace`, `off` (or {env-filter-syntax}[more complex filters,window=_blank])
 <3> String: E.g. `https://my-collector:4317` (Note: it must contain the scheme)
+
+== OpenTelemetry with Vector
+
+OpenTelemetry signals can be configured to behave like xref:logging.adoc[product logging]. Using the {vector-helm-chart}[vector helmChart] requires to configure `containerPorts` and `service.ports` as well as `customConfig.sources.otel`:
```
Techassi:
```suggestion
OpenTelemetry signals can be configured to behave like xref:logging.adoc[product logging]. Using the {vector-helm-chart}[Vector Helm Chart] requires `containerPorts`, `service.ports`, and `customConfig.sources.otel` to be configured:
```

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+    adapted to individual needs.
+<3> The OTLP `scope.name` contains the Rust module path (e.g.
+    `kube_runtime::controller`) and maps to the `.logger` field.
+<4> Log-record attributes (e.g. `built_info.*`) are collected into a single object
+    to keep the top-level schema flat.
+<5> Reference transformer output as sinks input.
+
+[NOTE]
+====
+Namespace information is currently not emitted by the operator
```
Techassi:
```suggestion
Namespace information is currently not emitted by the operator.
```

Techassi (commented):
Nicely done, just two small comments.

modules/concepts/pages/observability/telemetry.adoc
xeniape: What does "behave like product logging" mean exactly? A short explanation what that refers to, would be nice.

xeniape:
```suggestion
OpenTelemetry signals can be configured to behave like xref:logging.adoc[product logging]. Using the {vector-helm-chart}[Vector Helm Chart{external-link-icon}^] requires to configure the Helm values `containerPorts`, `service.ports`, and `customConfig.sources.otel`:
```

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+  sources:
+    otel:
+      type: opentelemetry
+      grpc:
+        address: 0.0.0.0:4317
+      http:
+        address: 0.0.0.0:4318
+----
+
+The `endpoint` in the operator's Helm values must point to the ports defined above.
```
xeniape:
```suggestion
The `endpoint` Helm values for the telemetry exporters in the Stackable Operator Helm Chart must point to the otel-grpc ports defined above.
```

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+
+[source,yaml]
+----
+telemetry:
+  otelLogExporter:
+    enabled: true
+    endpoint: http://vector-aggregator.<namespace>.svc.cluster.local:4317
```
xeniape, line 133: Actually, if there is a specific reason why this port was used and not otel-http for example, even though they are defined in Vector, feel free to elaborate why. If any of them can be used, maybe the above sentence can be adjusted to reflect that.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+  otelTraceExporter:
+    enabled: true
+    endpoint: http://vector-aggregator.<namespace>.svc.cluster.local:4317
+----
+
+=== Normalizing operator logs to the Stackable log schema
```
xeniape: Just a bit confused, above the sections starts with OpenTelemetry signals (but behaving like logs?), but here the subsection is about logs, so something different? So either that's something general for logs, so maybe needs to be a separate 2nd level section (==) or at least for me it needs a bit more of an explanation connecting the two concepts here.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+
+Operators do not run vector sidecars. Using OpenSearch, a VRL transform that maps the OTLP fields onto the
```
xeniape:
```suggestion
Stackable Operators do not run Vector sidecars. Using OpenSearch, a VRL transform that maps the OTLP fields onto the
```

xeniape: VRL? I think the abbrevation written out (or written out in parentheses next to it) would be more useful for people not knowing what that is (me :D)

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+xref:tutorials:logging-vector-aggregator.adoc#_watch_the_logs[Stackable log schema] is needed to avoid re-indexing:
+
+[source,yaml]
```
xeniape, line 143: Are these Helm values for Opensearch now? (Because of "using Opensearch" above) Maybe this also needs a linked Helm Chart then or mentioning that it's our own Opensearch Operator Helm Chart? And a clarification where to set the values.

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+        del(.dropped_attributes_count)
+  sinks:
+    opensearch:
+      inputs:
+        - vector
+        - normalize_otel_logs # <5>
+----
+
+<1> The `opentelemetry` source fans out into `otel.logs`, `otel.metrics`, and
+    `otel.traces`. Unconsumed sub-streams produce a startup warning.
```
xeniape:
```suggestion
<1> The `opentelemetry` source type diverges into `otel.logs`, `otel.metrics`, and
    `otel.traces` signals. Unconsumed sub-streams produce a startup warning.
```

(Don't know if "signals" is the right word, but some kind of naming would sound better I think. Oh, or is sub-stream the correct term? Then feel free to use that instead of signal in the suggestion)

modules/concepts/pages/observability/telemetry.adoc
```diff
@@ -93,3 +94,100 @@ telemetry:
...
+<2> Operators emit `service.name` (e.g. `stackable-airflow-operator`) as their
```
xeniape, line 183: Is `service.name` correct here? Or should it be `service_name` like in the line it references?

xeniape (changes requested):
Read it kind of "blind" as a non-expert, so maybe my confusions/questions could be used improve the documentation for other users not familiar with the topic. Keeping that in mind, my suggestions might not be the best, they are no blockers :)

modules/concepts/pages/observability/telemetry.adoc
xeniape: Adjusted the suggestion to include the external link indicator for the external Helm Chart

modules/concepts/pages/observability/telemetry.adoc
Maleware: It's `service.name` from the operator pov and only caught here as service_name as I don't want an attribute to the type service but rather something storing away that information

modules/concepts/pages/observability/telemetry.adoc
Maleware: The way I understood it, you can choose whatever port you want as long as you expose it in the aggregator.yaml shown above.

I understood that the gRPC standard port is 4317 and thus I kept the convention. Technically you could choose anything. I'm now not sure, I'd like to preserve some kind of convention and story throughout the guide. I think we could introduce `<gRPC-port>` and `<http-port>` if you like it better?

modules/concepts/pages/observability/telemetry.adoc
Maleware: Do you mean the following `customConfig` ?

If so, no. It's a map that defines how to map otel log fields so they fit the same schema as stackable products would look like.  Open search is just a very good example for a indexed backend. Since it's in our stack you are right that we might wanna hint it here. I'll fix it.

modules/concepts/pages/observability/telemetry.adoc
Maleware: Vector Remap Language. I shall do, thanks

modules/concepts/pages/observability/telemetry.adoc
Maleware: Adressed with [3c4d60f](https://github.com/stackabletech/documentation/pull/835/commits/3c4d60f62eacd6f28a80ae47b4b7dfa7e6e90f6a)

modules/concepts/pages/observability/telemetry.adoc
Maleware: Adressed with [3c4d60f](https://github.com/stackabletech/documentation/pull/835/commits/3c4d60f62eacd6f28a80ae47b4b7dfa7e6e90f6a)

modules/concepts/pages/observability/telemetry.adoc
Maleware: Good catch, hope it's better now: Adressed with [3c4d60f](https://github.com/stackabletech/documentation/pull/835/commits/3c4d60f62eacd6f28a80ae47b4b7dfa7e6e90f6a)

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: I think we should use the standard ports 4317 for gRPC, 4318 for HTTP. But yes they should be mentioned so the reader doesn't assume they are random.

modules/concepts/pages/observability/telemetry.adoc
NickLarsenNZ: Many injestors can't handle the dotted names. Some of them automatically remap them with `_`, others don't (I think OpenSearch Dataprepper didn't).

It's a mess out there, until everyone starts properly accepting Semantic Conventions :D

modules/concepts/pages/observability/telemetry.adoc
Maleware: Done [0241521](https://github.com/stackabletech/documentation/pull/835/commits/02415215f616703f59f61a62d6bbcabd1d644042)

modules/concepts/pages/observability/telemetry.adoc
Maleware: Donen [0241521](https://github.com/stackabletech/documentation/pull/835/commits/02415215f616703f59f61a62d6bbcabd1d644042)

Techassi (approved):

xeniape (approved):

PR #839, opened 2026-04-27
fix: Fix broken links and add redirect for external sites referencing old release notes link

xeniape:
Fix broken links and add a redirect to `netlify.toml` for external sites referencing the old URL with `.html` suffix


maltesander (approved):
LGTM!

PR #869, opened 2026-07-07
feat: Generate llms.txt for AI agents

lfrancke:
## Description

Adds an Antora extension that emits [`/llms.txt`](https://llmstxt.org/) at build time.
It gives AI agents a machine-readable entry point to the docs:

- indexes every page of the version that `stable` points to (the latest release; nightly in dev builds), grouped by module, with page descriptions from the `:description:` attribute
- each entry links the published page **and its AsciiDoc source on GitHub**, so an agent (or human) can go from any docs URL straight to the file to change
- cross-links the Stackable Hub's llms.txt for release/component/CRD data
- Only the stable version is indexed on purpose - agents should not land on the old docs

This is not perfect yet and we can do more later (e.g. a ".md" file next to each normal docs page so agents don't have to read all that HTML stuff" or a llms-full.txt which we might be able to use in the OpenSearch RAG demo) but that's separate.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -31,6 +31,7 @@ antora:
       mermaid_library_url: https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs
       script_stem: mermaid-scripts
     - ./supplemental-ui/lib/stackable-operator-helpers.js
+    - ./supplemental-ui/lib/llms-txt.js
 content:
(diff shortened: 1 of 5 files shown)


dervoeti (approved):
nit: some titles in llm.txt contain HTML entities, like `Configuration &amp; Environment Overrides`.
Just noticed it, should not be a problem for LLMs.

PR #878, opened 2026-07-14
chore: Add rate limiting to whole page

Techassi:
> [!WARNING]
> We don't actually know if aggregation by IP address is supported by our Netlify tier. So this change might needs to get reverted or adjusted.

This PR adds rate limiting to the complete page. Clients can request 25 times per 60 seconds. The limits are aggregated by client IP address.


StefanFl (approved):
lgtm

PR #880, opened 2026-07-15
chore: Add more granular rate limiting

Techassi:
This PR includes the following changes:

- Move rate limiting rules below other, more specific redirect rules
- Add more granular rate limiting rules grouped by
  - assets (250req/min)
  - pagefind assets (250req/min)
  - and a catch-all (50req/min)


StefanFl (approved):
lgtm

PR #882, opened 2026-07-15
chore: Only use 2 rate limit rules

Techassi:
Sadly, we can only have two rate limiting rules.

A follow-up PR will put the pagefind assets under the `/_/` path with all the other assets, so that the correct rate limiting rule will catch its assets.


StefanFl (approved):
lgtm

PR #883, opened 2026-07-15
chore: Place pagefind assets next to other assets

Techassi:
> [!NOTE]
> Wait for merge until the submodule has been updated.

This allows us to match requests to pagefind URLs using our asset rate limiting rule instead of the catch-all rule. The pagefind assets were previously /pagefind. Now, they are placed alongside other assets under /_/pagefind.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,6 @@
     "ui"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


StefanFl (dismissed):
lgtm

StefanFl (approved):
lgtm

PR #884, opened 2026-07-15
chore: Bump the normal rate limit back to 250

Techassi:
Netlify is doing something fucky. This basically immediately breaks normal browsing behaviour after 3/4 clicks while it clearly shouldn't.


StefanFl (approved):
good for the moment, let's remove them after bunny.net is active

PR #888, opened 2026-07-16
feat: Link the CRD reference on the Stackable Hub

lfrancke:
## Description

Our Hub [now](https://github.com/stackabletech/stackable-apps/pull/348) uses a similar URL style to the old CRD browser which means we can switch the docs to use the Hub instead of the old one.

To help with that there are new attributes in `antora.yml`:

- `crd-ref` - the listing (https://hub.stackable.tech/crds)
- `{crd-ref}/<group>/<apiVersion>/<kind>{crd-ref-release}` - a deep link
- `crd-ref-release` - `?release=dev` on main, the release number on release branches (same mechanism as `crd-docs-version` today)

Converted: the reference nav entry, the landing page link, and the five deep links (AuthenticationClass ×3, S3Bucket, S3Connection).

## Deliberately unchanged

- The old `crd-docs-*` attributes stay: the operator repos still build their deep links from them (~40 links). We need to update those separately :( (I'll do it...)
- `page-crd-docs` is removed - nothing in the UI templates reads it.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -30,6 +30,8 @@ asciidoc:
     # The CRD reference on the Stackable Hub. crd-ref links the listing;
     # deep links follow {crd-ref}/<group>/<apiVersion>/<kind>{crd-ref-release}
     crd-ref: "https://hub.stackable.tech/crds"
+    # Keep this in line with the 'version' key above: '?release=dev' on main,
+    # '?release=<major.minor>' on release branches (the release script sets it).
     crd-ref-release: "?release=dev"
     # DEPRECATED: the old CRD browser attributes below are still used by the
     # operator docs; they migrate to crd-ref repo by repo (operator sweep).
--- a/antora.yml
+++ b/antora.yml
@@ -27,10 +27,14 @@ asciidoc:
     # Whether this version is already end of life.
     # If true, a banner will be displayed informing the user.
     end-of-life: false
-    # use the attributes below to link to the CRD docs
+    # The CRD reference on the Stackable Hub. crd-ref links the listing;
+    # deep links follow {crd-ref}/<group>/<apiVersion>/<kind>{crd-ref-release}
+    crd-ref: "https://hub.stackable.tech/crds"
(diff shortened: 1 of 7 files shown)


antora.yml
```diff
@@ -27,10 +27,14 @@ asciidoc:
     # Whether this version is already end of life.
     # If true, a banner will be displayed informing the user.
     end-of-life: false
-    # use the attributes below to link to the CRD docs
+    # The CRD reference on the Stackable Hub. crd-ref links the listing;
+    # deep links follow {crd-ref}/<group>/<apiVersion>/<kind>{crd-ref-release}
+    crd-ref: "https://hub.stackable.tech/crds"
+    crd-ref-release: "?release=dev"
```
xeniape, line 35: Can we add a similar comment here as above `crd-docs-version` to not forget to update the version for release branches?

xeniape (changes requested):
Not in the changed files, so adding it here: https://github.com/stackabletech/documentation/blob/main/scripts/make-release-branch.sh#L121 also would need updating with the new value for the version

xeniape:
And just as a side note: The changes need to be backported to the older branches, otherwise the CRD links there would still point to the old url and not have the new values in `antora.yml`

antora.yml
lfrancke: Added in ab0c7a3, mirroring the crd-docs-version comment. With 577a588 the release script also sets it automatically.

lfrancke:
* make-release-branch.sh handled in https://github.com/stackabletech/documentation/commit/577a58852fedcc12603d8ca223e25c8db644caee

On backporting: Thanks... I always get confused by this, but you're right. This PR only covers a few "generic" links so that should actually be mostly fine but the "real links" are in each operator repo....so having to backport all of those to multiple branches will be painful. I'll see what I can do - I hope that maybe we can use some clever redirects instead.

xeniape (approved):

PR #890, opened 2026-07-16
perf: Replace Font Awesome with inline svg icons

lfrancke:
## Description

This builds on the work done in #889 

And it continues the work to improve our PageSpeed scores.

This is what I get on the preview:
image

And this is what I get on the live site, so quite an improvement (you can ignore the SEO score, due to the preview URL):

image

What tihs does in a nutshell:
We used to ship all of the Font Awesome package + eight webfont files for seven symbols we used from those. So that's about 400kB of stuff gone from every page load :partying_face: 

I let Claude inline those SVGs into one `icons.svg` file keeping proper attribution.

Drive-by: removes the `inlineDynamicImports` rollup option that Vite 8 warns is redundant.

The preview looks good to me. Look for the icons (external-link arrow in the top bar, content links, search icons, the social icons, ...). It all looks good to me.


lfrancke:
It looks big (22 files) but once you look at it you'll see it's mostly removals.

NickLarsenNZ (approved):
LGTM

PR #892, opened 2026-07-17
fix: Move pagefind assets back to /pagefind to fix broken search

dervoeti:
#883 moved the pagefind bundle from `/pagefind/` to `/_/pagefind/` so its requests would match the asset rate-limit rule. But pagefind derives its result baseUrl by stripping the `/pagefind/` suffix off the bundle path. With the bundle at `/_/pagefind/`, the derived baseUrl became `/_/`, which got prepended to every search result URL.

Rate limiting was removed entirely in #885, so #883 no longer serves any purpose. This reverts the bundle back to `/pagefind/`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "ui"
   ],
   "scripts": {
-    "build-search-index": "npx pagefind --site build/site --output-subdir _/pagefind"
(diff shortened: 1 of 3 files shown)


razvan (approved):
thanks

PR #895, opened 2026-07-21
docs: Add page descriptions to the Kubernetes, concepts and guides pages

lfrancke:
## Description

Adds `:description:` attributes to the 27 user-facing pages in this repo that had none (the Kubernetes distribution pages, the concepts maintenance/observability pages, the CRA page, the guides pages, the release guide).
The reason I added these is that I want them to show up in `llms.txt`.

Each description was written from the page content. I used AI to generate all of them, but I looked at them and think they are good enough.

Two deliberate omissions:

- `modules/concepts/pages/stackable_resource_requests.adoc` - it is an include target that also gets published as a page; a `:description:` there would leak into every including page (e.g. concepts/resources) and override their descriptions. Candidate for moving to `partials/` some day.
- The 39 contributor ADR pages - descriptions add nothing over their titles

The remaining description gaps live in the operator repos, stackablectl and demos.

This also removes the personas page which we never really used.


modules/ROOT/pages/kubernetes/eks.adoc
```diff
@@ -1,4 +1,5 @@
 = Amazon Elastic Kubernetes Service (EKS)
+:description: Run the Stackable Data Platform on Amazon Elastic Kubernetes Service (EKS). Make sure a default StorageClass is set so PVCs are provisioned.
```
Techassi: The "make sure..." part of the description seems a little off. It doesn't feel like it should be part of the page's description/overview.

modules/ROOT/pages/kubernetes/plusserver.adoc
```diff
@@ -1,3 +1,4 @@
 = plusserver Kubernetes as a Service
+:description: Notes on running the Stackable Data Platform on plusserver Kubernetes as a Service.
```
Techassi:
```suggestion
:description: Notes on running the Stackable Data Platform on plusserver's Kubernetes as a Service.
```

netlify.toml
```diff
@@ -7,6 +7,14 @@
   to = "https://hub.stackable.tech/crds"
   status = 301
 
+# The internal module is deleted on main but still exists on the release
+# branches; this blocks it for every published version without backports.
+[[redirects]]
+  from = "/home/:version/internal/*"
+  to = "/404.html"
```
Techassi: I would like to avoid redirecting to `.html` files as these redirects always cause issues. We should instead use:


```suggestion
  to = "/404"
```

Techassi (changes requested):
Most of the descriptions for pages under `pages/kubernetes` feel a little off. I like the ones in the style of "Notes on ...".

Others feel a too "tutorial-like". I also think that they shouldn't include details like "You need to do this" or "Custom stuff is needed because ...".

modules/ROOT/pages/kubernetes/eks.adoc
lfrancke: Agreed. I missed that.

netlify.toml
lfrancke: [`f589647` (this PR)](https://github.com/stackabletech/documentation/pull/895/commits/f5896474a0a8f80bd763c9979dcf0a2c79d9f382)

modules/ROOT/pages/kubernetes/plusserver.adoc
lfrancke: [`0d191a4` (this PR)](https://github.com/stackabletech/documentation/pull/895/commits/0d191a4aaec649ebe23cef9228255dc00a73bd65)

modules/ROOT/pages/kubernetes/eks.adoc
lfrancke: [`0d191a4` (this PR)](https://github.com/stackabletech/documentation/pull/895/commits/0d191a4aaec649ebe23cef9228255dc00a73bd65)

Techassi (approved):

PR #902, opened 2026-07-28
perf: Self-host mermaid and load it only on pages with diagrams

lfrancke:
## Description

Every page used to import 2MB of Mermaid JS from a CDN whether it contained a diagram or not.
Lighthouse reported this as "reduce unused JavaScript".

This PR changes it so Mermaid is a proper dependency (which also allows us to update it like other dependencies) and checks for a ".mermaid" element and if one exists it loads it on the fly.

Go to the listener operator installation page and check if the diagram renders: https://deploy-preview-902--stackable-docs.netlify.app/home/stable/listener-operator/installation/

It does for me :)

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -28,8 +28,6 @@ urls:
 antora:
   extensions:
     - require: '@sntke/antora-mermaid-extension'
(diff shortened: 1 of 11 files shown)


NickLarsenNZ (approved):
LGTM

PR #903, opened 2026-07-28
perf: Content-hash the bundle assets and take the icons from their package

lfrancke:
## Description

In earlier PRs I worked on PageSpeed where it complained about caching that isn't aggressive enough. We couldn't fully fix it then because our assets (JS, CSS etc.) all had static names (foo.css). So when we'd say "cache this for a day" we wouldn't see updates for a longish time.

This PR changes three things:

* Assets (CSSs & JS) are now content-hashed (foo-<hash>.css) so every time we change the content the hash changes and therefore we can cache them for a long time. The partials keep referring to the unhashed names and are being rewritten.
* In a [recent PR](https://github.com/stackabletech/documentation/pull/890) we introduced an `icons.svg` file with just the icons from FontAwesome we actually use, #911 then renamed that file. This PR changes it again to generate that file dynamically as well
	* This should finally move _all_ our dependencies into package.json and not be some special snowflake anymore
* Preloads `noto-sans-700` (Lighthouse showed it arriving late through the css chain on pages with bold text above the fold).


ui/build.mjs
```diff
@@ -97,7 +107,7 @@ await build(
         output: {
           assetFileNames: (asset) => {
             const name = asset.names[0] ?? '';
-            if (name === 'site.css') return 'css/site.css';
+            if (name === 'site.css') return 'css/site-[hash][extname]';
             if (/\.(woff2?|ttf)$/.test(name)) return 'font/[name][extname]';
             if (/\.(svg|png|gif|ico|jpg)$/.test(name)) return 'img/[name][extname]';
             return 'css/[name][extname]';
```
xeniape:
```suggestion
            return 'css/[name]-[hash][extname]';
```
Does it make sense to add the hash to the fallback also? Otherwise, if I understood it correctly, things not caught by the matches earlier will be cached for a year immutably?

ui/NOTICE
```diff
@@ -8,9 +8,11 @@ third-party components. Build-time minification strips embedded license
 header comments; this file preserves the attributions and is included
 in the published bundle.
 
-- Font Awesome Free 6.2.1 (https://fontawesome.com)
-  The inline svg icon path data is derived from Font Awesome Free icons
-  (CC BY 4.0). https://fontawesome.com/license/free
+- Font Awesome Free (https://fontawesome.com)
+  The icon sprite (img/icons.svg) is generated at build time from the
```
xeniape:
```suggestion
  The icon sprite (partials/icons.hbs) is generated at build time from the
```

ui/README.adoc
xeniape:
```suggestion
* `css`: Contains all the CSS. `site.css` contains imports of all the other files.
```

The last sentence can be dropped, since that's changed. (The preview of the suggestions looks like it's adding a line to me, it actually removes a sentence in that line)

xeniape (changes requested):
LGTM, just some minor things

lfrancke:
Thanks for the review. All of it made sense.
I applied all your suggestions.

xeniape (approved):

PR #904, opened 2026-07-28
perf: Defer the search UI, whole-card results, cleaner index

lfrancke:
## Description

This changes three things with the search:

* Another Google PageSpeed finding: The pagefind stuff was loaded on every page (~450ms render-blocking), now it's only loaded the first time you click on the search icon thing

* The thing that has annoyed me for years: The whole result card is now clickable and not only the title anymore

* Index hygiene: Added [`data-pagefind-ignore`](https://pagefind.app/docs/indexing/#removing-pages-from-pagefinds-index) attributes to the pagination nav and the EoL-banner

## Verification

Look at the preview...search something. Easy as that.

### DIFF
--- a/ui/src/partials/footer-scripts.hbs
+++ b/ui/src/partials/footer-scripts.hbs
@@ -2,57 +2,72 @@
 <script async src="{{{uiRootPath}}}/js/vendor/highlight.js"></script>
 <script async src="{{{uiRootPath}}}/js/vendor/tabs.js" data-sync-storage-key="preferred-tab"></script>
 <script>
-    // This script block loads the pagefind search UI.
+    // Loads the pagefind search UI on first use (see 07-search.js); the
+    // assets stay off the critical rendering path of every page.
     // Some styling is customized, these classes are found in search.css.
-    window.addEventListener('DOMContentLoaded', (event) => {
-        const instance = new PagefindModularUI.Instance({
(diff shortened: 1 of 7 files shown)


ui/src/partials/footer-scripts.hbs
```diff
@@ -2,57 +2,72 @@
 <script async src="{{{uiRootPath}}}/js/vendor/highlight.js"></script>
 <script async src="{{{uiRootPath}}}/js/vendor/tabs.js" data-sync-storage-key="preferred-tab"></script>
 <script>
-    // This script block loads the pagefind search UI.
+    // Loads the pagefind search UI on first use (see 07-search.js); the
+    // assets stay off the critical rendering path of every page.
     // Some styling is customized, these classes are found in search.css.
-    window.addEventListener('DOMContentLoaded', (event) => {
```
lfrancke, line 7: This looks worse than it is: It really is just different indentation mostly

xeniape (approved):
LGTM, what I noticed when trying out the preview (which is already like that on main), is that the search is not focusable when tabbing through the navigation. So keyboard users can't open it. This wasn't part of this PR, but just mentioning it if you want to add it.

lfrancke:
Thanks! I have more work planned in this area and I'll put it on the todo list.

PR #905, opened 2026-07-28
fix: Make netlify-fetch immune to stale cached release branches

lfrancke:
## Description

Production docs builds fail with `FATAL (antora): Could not find e9b0f5de... (branch: release-26.7)`.

### Root cause

e9b0f5de is not a commit, it is a blob: `ui/NOTICE` on release-26.7.
And it could not be found because (see the fix section below) we excluded `ui` from checkout because it used to be a submodule. And the [`isomorphic-git`](https://isomorphic-git.org/) used by Antora can't dynamically fetch missing content so.... it was missing.

### The fix

```
-             git -c gc.auto=0 checkout -q "$$branch" -- ':(exclude)ui' . ; \
+             git -c gc.auto=0 checkout -q "$$branch" -- . ; \
```

Do not exclude UI anymore... simple as that :(

### Separate preemptive fixes/hardening

These were "discovered" during the research...would not really help the root cause but good to have anyway.

- `gc.auto=0` for the loop commands, so background gc cannot race the ref updates.
- `set -e` + no `-q` on fetch: to just fail where the failure actually happens to get a better error message
- `git fetch --all --prune` to delete stale stuff from the cache, not only add or update

The first build after a new release branch can also fail because the branch list that is used uses the last cached git state from the previous run. That doesn't have the new release branch so it can fail. This now uses `git ls-remote --heads origin` (instead of `git branch -r`) to make sure we have the up-to-date state.

**Verified** by simulation (by Claude): blobless clone (like netlify's), stale `release-26.7` ref planted pointing at an outdated commit, `make netlify-fetch` run - the ref force-updates to the real remote tip, all content blobs of every release tip are present (only the deliberately excluded `ui/` blobs stay missing, which antora never reads), working tree clean, exit 0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,10 @@ netlify-fetch:
 	# race the ref updates and silently leave a stale local branch behind,
 	# which only surfaces much later as a missing-object error from antora.
 	# Failures abort the build immediately instead.
-	set -e; for branch in $$(git ls-remote --heads origin | sed 's|.*refs/heads/||' | grep -E 'release[/-]'); do \
+	set -eu; \
+	branches="$$(git ls-remote --heads origin | sed 's|.*refs/heads/||' | grep -E 'release[/-]')"; \
+	[ -n "$$branches" ] || { echo "netlify-fetch: no release branches from ls-remote - refusing to continue"; exit 1; }; \
+	for branch in $$branches; do \
 		git -c gc.auto=0 fetch origin "+refs/heads/$$branch:refs/heads/$$branch"; \
 		git -c gc.auto=0 checkout -q "$$branch" -- . ; \
 	done
--- a/Makefile
+++ b/Makefile
@@ -54,20 +54,25 @@ clean:
 	rm -rf cache
 
 # The netlify repo is checked out without any blobs. Antora needs local
-# release branches and their blobs in the object database, so this target
-# creates/updates the local branches and materializes their blobs via
-# pathspec checkouts - without ever switching HEAD (enabling branch previews!).
-# ui/ is excluded: release branches still reference it as a submodule and
-# its blobs are not needed (the UI bundle is built from the current checkout).
+# release branches and their complete tip trees in the object database (it
+# reads them with isomorphic-git, which cannot fetch missing objects on
+# demand), so this target creates/updates the local branches and
+# materializes their blobs via pathspec checkouts - without ever switching
+# HEAD (enabling branch previews!).
 netlify-fetch:
 	# netlify messes with some files, restore everything to how it was:
 	git reset --hard
 	# fetch, because netlify does caching and we want to get the latest commits
-	git fetch --all
-	for remote in $(shell git branch -r | grep -E 'release[/-]'); do \
-		branch="$${remote#origin/}"; \
-		git fetch -q origin "+refs/heads/$$branch:refs/heads/$$branch"; \
-		git checkout -q "$$branch" -- ':(exclude)ui' . ; \
+	git fetch --all --prune
+	# The branch list comes from the remote so it is current even on the first
+	# build after a new release branch appears (the cached remote-tracking refs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -63,11 +63,16 @@ netlify-fetch:
...
-		git fetch -q origin "+refs/heads/$$branch:refs/heads/$$branch"; \
-		git checkout -q "$$branch" -- ':(exclude)ui' . ; \
+	git fetch --all --prune
+	# The branch list comes from the remote so it is current even on the first
+	# build after a new release branch appears (the cached remote-tracking refs
+	# lag behind). Auto-gc is disabled inside the loop: a background gc can
+	# race the ref updates and silently leave a stale local branch behind,
+	# which only surfaces much later as a missing-object error from antora.
+	# Failures abort the build immediately instead.
+	set -e; for branch in $$(git ls-remote --heads origin | sed 's|.*refs/heads/||' | grep -E 'release[/-]'); do \
```
NickLarsenNZ: `set -e` should probably be right at the top to catch any error.

If bash, then this is often handy:

```
set -euo pipefail
```

if plain `sh`, then:

```
set -eu
```

NickLarsenNZ (dismissed):
Left a comment, but in any case, it's an improvement.

Makefile
lfrancke: [`04f2864` (this PR)](https://github.com/stackabletech/documentation/pull/905/changes/04f28648e8dbe4ad19123abbf462a54f8b4405e9)

NickLarsenNZ: I mean right at the top of the script. It's essentially "strict mode" for the whole script.

NickLarsenNZ: ie:

```Makefile
netlify-fetch:
  set -eu
  # netlify messes with some files, restore everything to how it was:
  git reset --hard
  # fetch, because netlify does caching and we want to get the latest commits
  git fetch --all --prune
```

No semicolon or `\` needed.

NickLarsenNZ: Ahh sorry, my bad. It's a Makefile :facepalm:

NickLarsenNZ: It's good as-it now.

NickLarsenNZ: TIL `.ONESHELL:`

See: https://seanbone.ch/makefile-hacks/#Running_all_lines_of_a_target_in_the_same_shell

Techassi (commented):
This is git dark magic, but it looks _okayish_ (for me personally).

I'm unsure why we need this complicated fetch `"+refs/heads/$$branch:refs/heads/$$branch"` instead of `$$branch`, but again, dark magic.

Techassi:
I just spotted this

```
6:02:22 PM: branches="$(git ls-remote --heads origin | sed 's|.*refs/heads/||' | grep -E 'release[/-]')"; \" build_id=6a68d27df740ba0008ae6073 cache_from=
6:02:22 PM: \tgit -c gc.auto=0 checkout -q "$branch" -- . ; \" firebase_kafka_delivery_enabled=true repo_fetch_ref=pull/905/head cache_from=
6:02:22 PM: done
```

in the deploy log. What is `firebase_kafka_delivery_enabled=true` (or more generally speaking all of these key-value pairs)?.

lfrancke:
@Techassi: 
This to me is dark magic as well and I am going to trust AI here.
> The explanation it gives me is this: `git fetch origin <branch>` only updates `FETCH_HEAD`, it doesn't create or update the local branch ref. Antora reads the release branches as local branches (refs/heads/…) via isomorphic-git, so we need the explicit refspec to materialise them without checking them out. The + forces the update so a stale ref restored from Netlify's build cache gets overwritten instead of failing the fetch.

And I believe that stale-cache thing is exactly what this PR is guarding against.

On the log stuff:  Googling "firebase_kafka_delivery_enabled" leads to zero results :o
But I believe I've seen it before and I think it must be some Netlify internal stuff.

Techassi (approved):

PR #906, opened 2026-07-29
chore: Sync the vendored UI with upstream antora-ui-default, part 1: clean ports

lfrancke:
We based our UI on the [Antora default UI](https://gitlab.com/antora/antora-ui-default) but have not caught up with fixes that have landed there for four or so years :partying_face: 

So, this is PR basically takes a lot of the missing commits (some come later in separate PRs due to conflicts etc.) and integrates them here.

Most of these are just "fixes" for stuff we don't see, but there are a few things that are user-visible here:

1. Smooth scrolling: Clicking an entry in the right-hand ["on this page" TOC](https://deploy-preview-906--stackable-docs.netlify.app/home/stable/hdfs/#_supported_versions) or any in-page anchor link now animates to the target instead of jumping (loading a URL with a #fragment still positions instantly).
2. Copy-to-clipboard: The "copied" toast fades faster, and stacking fixes mean the copy button/toast can no longer be overlapped by other elements on [code blocks](https://deploy-preview-906--stackable-docs.netlify.app/home/stable/hdfs/getting_started/installation/). 
3. Description-list spacing: the margin around [descriptions](https://deploy-preview-906--stackable-docs.netlify.app/home/stable/commons-operator/restarter/) is slightly tidier.
4. tab-size: 4 code samples containing literal tab characters render 4 wide instead of 8, not sure if we have this anywhere

We skipped a few upstream commits which are relevant only for other fonts or styles where we diverged on purpose. Some of that will come in later PRs.

> [!IMPORTANT]
> I'd like a review on this but I'll need to force merge anyway because of the (obviously) missing CLAs. There is a note about this in this PR


xeniape (changes requested):
LGTM, one minor thing:
The commits here change the styling a bit, we override some of those in `ui/src/css/admonition.css` for example. There are now two dead code places in there due to the changes: lines 13-15 and lines 34-36.
This breaks the styling for them, example here: https://deploy-preview-906--stackable-docs.netlify.app/home/stable/contributor/code-style-guide/#_cargo_toml (the green/red boxes).

lfrancke:
How the hell did you find that? But thanks for catching it. It took me two minutes staring at the screen to see the difference. But I hope it's fixed now.

xeniape (approved):

PR #910, opened 2026-08-03
chore: add known issue to 26.7 release

sweb:
Add Spark image issue to known issues for 26.7


NickLarsenNZ (dismissed):
LGTM

Techassi (dismissed):
We technically don't need this to be a list item, but I'm happy to leave it as is if there will be more items down the line.

Techassi (approved):

PR #911, opened 2026-08-04
fix: display icons on local builds

xeniape:
The following description is provided by Claude:

## Problem

  Icons were invisible in local builds. Opening a generated page directly from disk showed no
  search icon, no footer social icons and no "external link" arrows next to outbound links.

  Separately, the link icons on the glossary page are missing on the live site for all released
  versions (e.g. /home/stable/reference/glossary/), while nightly is fine.

  ## Cause

  Both problems trace back to #890, which replaced the icon font with SVG icons.

  The icons now live in a single shared file that each page points at. That pointer is a request
  for a second file, and browsers refuse those on pages opened from disk — so in local builds
  every icon silently disappeared. Served over HTTP it works, which is why the live site looked
  fine and this slipped through review.

  The glossary is a different case. Its icon is defined in the page source itself, which is
  versioned, so released branches still contain the old icon-font markup. The stylesheet that
  made it visible was removed in #890, so those icons now render as nothing.

  ## Approach

  The icon definitions are now included directly in every page, and icons refer to them within
  the same page. Nothing needs to be fetched, so they work identically whether the site is opened
  from disk, served locally, or deployed.

  This adds roughly 2.7 kB (compressed) per page. The far larger saving from #890 — dropping
  Font Awesome and its web fonts — is unaffected.

  ## Not covered here

  The glossary icons remain broken in already-released versions, since the fix is in page source
  that those branches don't have. -> This would need separate PRs backporting the change

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -327,7 +327,7 @@ asciidoc:
   attributes:
     base-repo: https://github.com/stackabletech
     # an external icon. Create a link opening in a new tab like this: http://my-url.com[Text {external-link-icon}^]
-    external-link-icon: "<svg class=\"svg-icon\" style=\"width: 0.75em; height: 0.75em; margin-left: 3px;\" aria-hidden=\"true\"><use href=\"/_/img/icons.svg#icon-external-link\"/></svg>"
+    external-link-icon: "<svg class=\"svg-icon\" style=\"width: 0.75em; height: 0.75em; margin-left: 3px;\" aria-hidden=\"true\"><use href=\"#icon-external-link\"/></svg>"
 # the default caching directory is ./.cache/antora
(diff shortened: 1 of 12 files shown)


NickLarsenNZ:
Using the deploy preview, I see it works on nightly, but will it work for the older releases?

Nightly:

image

26.7:

image

xeniape:
> Using the deploy preview, I see it works on nightly, but will it work for the older releases?

That's one of the things mentioned, that are currently broken (due to #890), and that is not addressed by this PR. It needs changes in the release branches, unrelated to this PR (the one-line change would just look sligthly different pre or post this PR).

NickLarsenNZ (approved):
I can't say I understand each of the changes, but it looks fine from my quick check on the preview site.

PR #919, opened 2026-08-12
perf: Give the navbar and footer logos explicit dimensions

lfrancke:
If you include images but they are not loaded yet the browser doesn't know how much "space" to reserve for them. So they do whatever until the picture is there and then they have to "reflow" when the image arrives.

One way around this is to give images an explicit size in HTML so the browsers know how much space to reserve. This PR does that for two occasions. This was again flagged by Lighthouse.

### DIFF
--- a/ui/src/partials/header.hbs
+++ b/ui/src/partials/header.hbs
@@ -2,7 +2,7 @@
   <nav class="navbar">
     <div id="topbar-left" class="navbar-brand">
       <a class="navbar-item" href="{{{ relativize (versioned "home" page "index.html") }}}">
-        <img src="{{{uiRootPath}}}/img/stackable-logo.png">
+        <img src="{{{uiRootPath}}}/img/stackable-logo.png" width="199" height="30">
       </a>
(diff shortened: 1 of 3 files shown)


ui/src/partials/header.hbs
```diff
@@ -2,7 +2,7 @@
   <nav class="navbar">
     <div id="topbar-left" class="navbar-brand">
       <a class="navbar-item" href="{{{ relativize (versioned "home" page "index.html") }}}">
-        <img src="{{{uiRootPath}}}/img/stackable-logo.png">
+        <img src="{{{uiRootPath}}}/img/stackable-logo.png" width="199" height="30">
```
xeniape, line 5: Image

That looks weirdly stretched

xeniape (changes requested):

lfrancke:
Thanks again for catching that. I don't know what I looked at. This should be fixed now.
I'm still a bit annoyed that we have two different PNGs and a SVG in here but that's for another PR...

xeniape (approved):

PR #920, opened 2026-08-12
chore: Sync the vendored UI with upstream antora-ui-default, part 2: JS and a11y ports

lfrancke:
Part 2 of the vendored-UI re-sync with upstream, following #906

This PR focuses on the JS and accessibility stuff (after this we have CSS left)

Similar to last time: CLA checks will fail by design so I'll manually merge this but still want a review. See #906 for details.

## User-visible

One new feature and it's not very exciting. The ToC that appears on [content pages ](https://deploy-preview-920--stackable-docs.netlify.app/home/stable/druid/) now has a expand/collabse button on hover.

The rest is mostly bugs for edge cases (switching from mobile to desktop view, a console warning and so on).

## Four upstream commits did not land

> [!NOTE]
> This section is entirely Claude generated and can be ignored. I left it in as a reference should we ever wonder why a certain commit isn't included. 

- `fb151724` (overscroll) — already present in all four affected places.
- `2bcdc3a2` (nav resilience) — already applied; `git am` detected it.
- `9c6d2489` (highlight.js `none` language) — superseded by `da14ed68`, which is in this PR. Untagged blocks are no longer passed to highlight.js at all, so there is no unknown-language warning left to fix. Upstream made the same move itself, from defining `none` in 2022-03 to skipping untagged blocks in 2022-07.
- `cdd7b764` (scrollbar normalisation) — deferred. It needs two colour variables that do not exist yet (`--scrollbar-track-color`, `--scrollbar_hover-thumb-color`), and picking values is a design decision in a file the planned token work will rewrite. The `@media (pointer: fine)` gating in it is worth taking separately.


ui/src/css/header.css
```diff
@@ -3,8 +3,10 @@
  * It also handles the mobile version of the menu.
  */
 
-html.is-clipped--navbar {
-  overflow-y: hidden;
+@media screen and (max-width: 1023.5px) {
```
xeniape:
```suggestion
@media screen and (max-width: 1200px) {
```

Our navbar breakpoint is at 1200px (see line 288), so the styling should also apply until that width

xeniape (changes requested):
LGTM, just one minor thing

lfrancke:
Thanks for the review and the fix!

PR #923, opened 2026-08-13
feat: Read supported versions from the Hub, including nightly

lfrancke:
This changes the documentation to NOT need the supported-version partials to be maintained manually anymore.
At build time it connects to the Stackable Hub to get the versions and writes the partials itself.

For unreleased versions (nightly) it says "not decided yet" (in nicer words).

See:

* https://deploy-preview-923--stackable-docs.netlify.app/home/stable/operators/supported_versions/
* https://deploy-preview-923--stackable-docs.netlify.app/home/nightly/operators/supported_versions/
* https://deploy-preview-923--stackable-docs.netlify.app/home/nightly/hbase/#_supported_versions
* https://deploy-preview-923--stackable-docs.netlify.app/home/stable/hbase/#_supported_versions

> [!NOTE]
> When this is merged I'll go ahead and open PRs to delete the partial from every operator and then I assume some of the issue templates for version bumps need to be adjusted.

### DIFF
--- a/lib/hub-supported-versions.js
+++ b/lib/hub-supported-versions.js
@@ -21,6 +21,7 @@ const ospath = require('path')
 
 const HUB_API = 'https://hub.stackable.tech/api/v1/components'
 const PARTIAL = 'supported-versions.adoc'
+const HUB_TIMEOUT_MS = 10_000
 
 // docs module name -> Hub component slug
 const MODULE_TO_SLUG = {
@@ -70,8 +71,10 @@ module.exports.register = function () {
           component: 'home', version, module: moduleName, family: 'page', relative: 'index.adoc',
         })) continue
 
-        const body = renderPartial({ components, slug, version, logger })
-        if (!body) continue // no data and a repo copy is present: leave it alone
+        const body = renderPartial({
+          components, slug, version, logger, hasRepoCopy: Boolean(existing),
+        })
+        if (!body) continue // the repo ships a copy and the Hub has nothing better
 
         if (existing) {
           existing.contents = Buffer.from(body, 'utf8')
@@ -89,7 +92,7 @@ module.exports.register = function () {
 }
 
 // Returns the AsciiDoc body, or undefined to mean "leave whatever is there".
-function renderPartial ({ components, slug, version, logger }) {
+function renderPartial ({ components, slug, version, logger, hasRepoCopy }) {
   const header = `// Generated at build time from ${HUB_API}/${slug}.\n` +
     '// Do not edit: the Portal is the source of truth. See lib/hub-supported-versions.js.\n'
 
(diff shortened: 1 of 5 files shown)


lib/hub-supported-versions.js
```diff
@@ -0,0 +1,194 @@
...
+  superset: 'superset',
+  trino: 'trino',
+  zookeeper: 'zookeeper',
+}
+
+const STATUS_SUFFIX = {
+  lts: ' (LTS)',
+  deprecated: ' (deprecated)',
+  experimental: ' (experimental)',
+  preview: ' (preview)',
```
xeniape, line 47: Maybe I'm just seeing it for the first time, but do we actually use that one and how is it different from experimental?

lib/hub-supported-versions.js
```diff
@@ -0,0 +1,194 @@
...
+    const cacheDir = ospath.join(playbook.dir || '.', playbook.runtime.cacheDir || './cache', 'hub')
+    const components = await fetchComponents(cacheDir, logger)
+
+    const component = contentCatalog.getComponent('home')
+    if (!component) return logger.info('no home component, nothing to do')
+
+    let written = 0
+    for (const { version } of component.versions) {
+      for (const [moduleName, slug] of Object.entries(MODULE_TO_SLUG)) {
+        const existing = contentCatalog.getById({
```
xeniape, line 65: I would like to rename that variable, "existing" isn't descriptive enough, what exists? Based on the call it should return the versions file, so maybe "versionsPartial" or "existingVersionsPartial"/"existingPartial"?

lib/hub-supported-versions.js
```diff
@@ -0,0 +1,194 @@
...
+        // A module we do not carry in this docs version at all: nothing includes
+        // the partial, so do not invent one.
+        if (!existing && !contentCatalog.getById({
+          component: 'home', version, module: moduleName, family: 'page', relative: 'index.adoc',
+        })) continue
+
+        const body = renderPartial({
+          components, slug, version, logger, hasRepoCopy: Boolean(existing),
+        })
+        if (!body) continue // the repo ships a copy and the Hub has nothing better
```
xeniape, line 77: I kind of don't understand the whole decision tree here. So for older branches where the versions partial exists, it still fetches the Hub information and overrides. Why not just keep the existing file in the older branches instead? The information there shouldn't change. It would also simplify the code, because that whole `existing` variable usage wouldn't be needed. (If I understood that correctly `renderPartial` itself also decides to skip the override if there is no related hub release, not sure when that is the case, seems no-op to me if existing file is used earlier). But maybe I misunderstand something.

xeniape (changes requested):
I stopped the review for now because I didn't understand the business logic, will continue once that's answered.

Also some things I noticed, not sure if they were deliberate:
- The Hive 4 issues section disappears in this PR version
- The Druid Note also disappears

lib/hub-supported-versions.js
lfrancke: No, you're right.
This was dumb. It came from https://hub.stackable.tech/llms.txt and I just never thought about it. I'll fix it.

lib/hub-supported-versions.js
lfrancke: Agreed. I'll rename it.

lib/hub-supported-versions.js
lfrancke: Okay, this probably came from my poor understanding of how exactly the docs work especially with older branches and me then telling Claude the wrong things.

I did look at a few of the version pages (see my links above) but obviously not all of them. I missed that they are not all the same. I thought they are all "just lists".

Let me think about this a bit more because a) I want to get away from manually dealing with these lists b) some of that "manual" information is useful (the Hive stuff, Druid deprecation) and I just don't know where else it should live yet and c) _some_ of the old information is actually wrong and/or at least disagrees with the Hub. (e.g. HDFS 3.5 is marked as LTS in the docs while we wanted to keep it at 3.4 - https://docs.stackable.tech/home/stable/hdfs/)

lfrancke:
Thank you for the review. That requires a bit more thinking on whether I e.g. want to allow comments for versions and releases in the Hub. I'll park this for a bit to think about it.

PR #925, opened 2026-08-20
ci: Harden workflows

Saul-STFC:
This PR scopes the permissions of the workflow and bumps dep version hashes


.github/workflows/pr_pre-commit.yml
```diff
@@ -4,14 +4,19 @@ name: pre-commit
...
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 0
-      - uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+          persist-credentials: false
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           python-version: '3.12'
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
```
Techassi: We should switch to our own ready-to-use `run-prek` action here. But that can be done in a follow-up PR.

Techassi (dismissed):

.github/workflows/pr_pre-commit.yml
Techassi: Done in f5decd41.

Techassi (dismissed):
Approved.

CC'ing @dervoeti for a second pair of eyes.

dervoeti (approved):
@Techassi I fixed two small things in https://github.com/stackabletech/documentation/pull/925/commits/96d86727916e1c25a4849b31c7b22c7cc9a875ef
LGTM now

PR #928, opened 2026-09-03
fix: Lifecycle Policies menu entry not correct

StefanFl:
The entry for the page "Lifecycle policies" in the top menu was "Policies", which doesn't reflect the actual content.


xeniape (approved):

xeniape:
Not sure why it does not show the change in the preview, but we'll see in the live version

PR #929, opened 2026-09-03
fix: Make keyboard focus visible and name the unnamed controls

lfrancke:
This continues my journey to improve our PageSpeed score.
We are not "great" on accessibility yet and this tries to fix some of that. It goes from 83 to 93 for me with this PR and more will come.

New:
Homepage
image

Operator
image

Old:
Homepage
image

Operator
image

There should be _no_ user visible changes - I clicked around and it all looked good. But keyboard focus works much better. The menu bar for example can now be tabbed through and you can reach the search thing.
In the sidebar menu it is now also possible to tab through everything. Some items also didn't have a visible "ring" on it when selected via keyboard. That is all fixed. And some icons etc. now have alt names.

There will be more to come.


NickLarsenNZ (approved):
LGTM

PR #931, opened 2026-09-03
fix: Lifecycle Policies menu entry not correct (now in header menu)

StefanFl:
Not sure what the file edited in #928 is for, but the header menu will get the proper title for Lifecycle Policies with this PR.


xeniape (approved):
Ah! Yes, the earlier PR "fixed" it in the navigation bar to the left, where it was already correct, because if there is nothing specified in `[]` it takes the title of the page I think, which was already "Lifecycle Policies". So we could remove it again if the title ever changes so it's reflected in the navigation bar (which I guess is unlikely so it does not hurt either)

PR #932, opened 2026-09-04
fix: Align the footer social links with the other Stackable sites 

lfrancke:
The title is a lie. It doesn't align with anything but I think it now covers everything from Hub & the homepage in a consistent way. So I believe this could be the new footer for all of them.


maltesander (approved):
LGTM!

